PR #2, opened 2019-09-25
CO-1596: Add swagger spec

jemisonf:


openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
-            $ref: '#/definitions/PetResult'
-        '404':
-          description: PetResource not found
+            $ref: "#/definitions/IngredientsResult"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /pizzas/{pizzaID}/ingredients/{ingredientID}:
+    patch:
```
jemisonf: not totally clear if `put` or `patch` is the right method to use here -- I went with `patch` because I was thinking in terms of updating the pizza resource

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+        200:
+          description: "A single dough recipe"
+          schema:
+            $ref: "#/definitions/DoughResult"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /pizzas/{pizzaID}/dough/{doughID}:
+    patch:
```
jemisonf: similar to comment on line 172 -- could potentially be `put` as well I think in this case

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
-        - pets
+        - pizza
       parameters:
         - in: body
-          name: body
+          name: attributes
+          description: attributes to assign to the new pizza
+          required: true
+          schema:
+            $ref: "#/definitions/PizzaAttributes"
```
jemisonf: I used this pattern in a couple other places -- `post`ing a `__Attributes` object -- and I'm not sure if this is an accepted pattern for this or not

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
           schema:
-            $ref: '#/definitions/PetResult'
-        '404':
-          description: PetResource not found
+            $ref: "#/definitions/IngredientsResult"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /pizzas/{pizzaID}/ingredients/{ingredientID}:
```
jemisonf: this (and similar with `/pizzas/{pizzaID/doughs/{doughID}` seemed like the most obvious and simplest way to implement this, but not necessarily the best? reminds me of the conversation about the best way to create/update people's addresses in the persons API

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+        type: array
+        items:
+          $ref: "#/definitions/ErrorObject"
   ErrorObject:
     properties:
       status:
         type: string
         description: HTTP status code
-        example: '123'
+        example: "123"
```
jemisonf: auto-corrected to double quotes by my linter, could probably change this is it's an issue

openapi.yaml
tsoliangwu0130: Interesting topic. Here're some questions I'd suggest you think about first:

1. Will the `ingredientID` be global unique? In other words, will `/pizzas/1/ingredients/A` and `/pizzas/2/ingredients/A` be the same ingredient or not?
2. How aobut if I want to add a new ingredient to pizza 1? Looks like you didn't have a `POST` method for `/pizzas/{pizzaID}/ingredients`. Correct me if I'm wrong, but I guess the ID you designed should be an auto-increment or a random string which will be generated by the database. If your API do support `POST` for this endpoint, does that mean you'll create repeating records for the same ingredient? In other words, if I want to add `pineapple` to both pizza 1 and pizza 2, I'll need to send two POST requests to `/pizzas/1/ingredients` and `/pizzas/2/ingredients` right? In this case, the auto-generated ingredients won't necessarily be the same. How'd you design your database schema to handle this?
3. If we PATCH the pineapple for pizza 1, will the pineapple in pizza 2 be changed as well? Same thing for the `DELETE`.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
-       $ref: '#/definitions/ErrorResult'
-securityDefinitions:
-  OAuth2:
-    type: oauth2
-    flow: application
-    tokenUrl: https://api.oregonstate.edu/oauth2/token
-    scopes:
-      full: Full access to the API
+            type: boolean
+            example: true
```
tsoliangwu0130: We can't just return a boolean as a result. It must be a JSON object.

Ref: https://jsonapi.org/format/#crud-updating
> The PATCH request MUST include a single resource object as primary data. The resource object MUST contain type and id members.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+        200:
+          description: "Returns true is the resource was deleted"
+          schema:
+            type: boolean
+            example: true
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /pizza/{pizzaID}/dough:
```
tsoliangwu0130: Same comment as `ingredients`.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+          description: Search doughs by name
+        - in: query
+          name: gramsFlour
+          required: false
+          type: integer
+          description: Search doughs by grams flour used
+        - in: query
+          name: gramsWater
+          required: false
+          type: integer
```
tsoliangwu0130: Same comment as `ovenTemp` in `/pizzas` endpoint. In addition, should this be a floating-point number? Same thing for other parameters.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+          description: Search doughs by proof time
+        - in: query
+          name: waterTemp
+          required: false
+          type: integer
+          description: Search doughs by water temperature at mix time
+        - in: query
+          name: flourType
```
tsoliangwu0130: How many possible flour types will it be? Should this be an `enum` parameter?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+          name: attributes
+          description: attributes of the new dough recipe
+          schema:
+            $ref: "#/definitions/DoughAttributes"
+      responses:
+        200:
+          description: "The created dough recipe"
+          schema:
+            $ref: "#/definitions/DoughResult"
+        500:
```
tsoliangwu0130: Besides `200` and `500`, can you think of any possible status codes?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+      responses:
+        200:
+          description: "The updated dough recipe"
+          schema:
+            $ref: "#/definitions/DoughResult"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+      deprecated: false
```
tsoliangwu0130: You don't need to specify it if it's not deprecated.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+  /doughs/{doughID}/pizzas:
```
tsoliangwu0130: This endpoint makes me wonder can't we also return dough data when querying `/pizzas`? In other words, what's the main difference between `/doughs/A/pizzas` and `/pizzas?doughID=A`?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+      responses:
+        200:
+          description: "A list of pizzas"
+          schema:
+            $ref: "#/definitions/PizzasResults"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /ingredients:
```
tsoliangwu0130: See my comment in `/doughs` endpoint.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+      data:
+        $ref: "#/definitions/PizzaResource"
+  PizzaResource:
+    properties:
+      id:
         type: string
-        format: url
-        description: Self-link of current resource
-  PaginationLinks:
+        description: a unique ID of a pizza
```
tsoliangwu0130: If possible, I'd recommend giving an example for each property, it'll be super helpful for whoever use this API. Example: https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L391

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
-        description: The first page of data
-      last:
+        description: the name of the pizza
+        example: My Favorite Margherita
+      ovenTemp:
+        type: integer
+        description: the temperature to bake the pizza at, in Fahrenheit
+        example: 500
+      bakeTime:
+        type: integer
```
tsoliangwu0130: Should this be a floating-points number?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+      responses:
+        200:
+          description: "The updated ingredient"
+          schema:
+            $ref: "#/definitions/IngredientResult"
+        404:
+          $ref: "#/responses/404"
+        500:
+          $ref: "#/responses/500"
+  /ingredients/{ingredientID}/pizzas:
```
tsoliangwu0130: How if we want to sure all pizzas made with more than one ingredients? For example, how to search all pizzas contain `pineapple` and `mango` (😨) ?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
       parameters:
         - in: body
-          name: body
+          name: attributes
+          description: attributes to assign to the new pizza
+          required: true
+          schema:
+            $ref: "#/definitions/PizzaAttributes"
+      responses:
+        200:
```
tsoliangwu0130: Based on [JSON API - create](https://jsonapi.org/format/#crud-creating-responses-201), if the resource has been successfully created, it should return a `201`.

tsoliangwu0130:
Can you also add the JIRA ticket number to this PR title?

openapi.yaml
jemisonf: `ingredientID` is a global unique value, with the idea that you could theoretically do something like finding all pizzas that have pepperoni as an ingredient.  So the workflow to add a new ingredient to a pizza is:

1. If the ingredient doesn't already exist, `POST /ingredients` with the name and type of the new ingredient
2. Add the ingredient to the pizza with `PATCH /pizzas/{pizzaID}/ingredients/{ingredientID}`

If you update the ingredient you'd use the `PATCH ingredients/{ingredientID}` endpoint and not the `PATCH /pizzas/{pizzaID}/ingredients/{ingredientID}`, which is only for adding ingredients to pizzas. `DELETE /pizzas/{pizzaID}/ingredients/{ingredientID}` only removes the ingredient from the pizza. As it stands, you can't delete ingredients from the database.

openapi.yaml
jemisonf: That sounds like a better approach since it would let you compose that with other filters. I think I'll remove the `/doughs/{doughID}/pizzas` and replace it with the parameter approach.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
-          description: Search by pet species
+          name: name
+          description: Search pizzas by name
+          required: false
+          type: string
+        - in: query
+          name: ovenTemp
+          description: Search pizzas by bake temperature (in degrees Fahrenheit)
           required: false
           type: string
```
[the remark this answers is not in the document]
jemisonf: It would return only pizzas with that exact oven temp. That should definitely be a number.

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+          description: "A list of pizzas"
           schema:
-            $ref: '#/definitions/PetsResult'
-        '400':
-          $ref: '#/responses/400'
-        '500':
-          $ref: '#/responses/500'
+            $ref: "#/definitions/PizzasResults"
+        500:
+          $ref: "#/responses/500"
```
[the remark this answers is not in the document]
jemisonf: Sounds like [400: bad request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
           schema:
-            $ref: '#/definitions/PetResult'
-        '400':
-          $ref: '#/responses/400'
-        '409':
-          description: Conflict
+            $ref: "#/definitions/PizzaResult"
+        404:
+          $ref: "#/responses/404"
+        500:
```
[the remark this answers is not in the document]
jemisonf: Also 400?

openapi.yaml
```diff
@@ -25,254 +37,677 @@ security:
...
+      tags:
+        - pizza
+      parameters:
+        - in: path
+          name: pizzaID
+          required: true
+          type: string
+          description: "ID of a pizza to retrive"
+      responses:
+        200:
```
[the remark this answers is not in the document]
jemisonf: That makes sense 👍

openapi.yaml
jemisonf: My sense is integer is fine; I've never seen fractions of grams used for these measurements since most kitchen scales aren't precise enough for it to really matter.

openapi.yaml
jemisonf: My goal here was to allow people to be fairly granular in their specifications about flour types, so an enum might be too restrictive. I think the ideal method here would be to have another `flour` resource but that would add a lot of additional complexity to the API.

openapi.yaml
jemisonf: This would also make sense to have a `400` response

openapi.yaml
jemisonf: I would hope no one's searching for `pineapple` and `mango` pizzas, but if they were we could do the same thing you suggested with the `doughId` and move it to a parameter that would allow you to pass an array of ingredients

openapi.yaml
jemisonf: Sounds good, I will adjust these to use the `XXXResource` object instead

openapi.yaml
jemisonf: If I'm reading [this](https://jsonapi.org/format/#crud-updating-responses-204) correctly, an update with no side effects can just return a `204` right?

openapi.yaml
jemisonf: Similar to the other case, I've never seen fractional/decimal measurements used for temperatures in recipes

jemisonf:
Making a number of changes:
* fixing syntax, logic errors
* adding `400` responses to `PATCH` and `POST` methods
* changing `200` to `204` responses in `PATCH` and `DELETE` methods
* removing quotes around strings basically everyone and adding them to the response code keys
* downcasing the 'd' in "ID"
* pulling pizzaID parameter to `parameters` definition
* changing `PATCH` body parameters to match openAPI specs
* removing `GET XXX/{pizzaID}`endpoints and replacing with parameters in `GET /pizzas`
* add `example`s where missing

openapi.yaml
```diff
@@ -25,105 +32,433 @@ security:
...
-      description: Get all pets
-      operationId: getPets
+        - pizza
       parameters:
-        - $ref: '#/parameters/pageNumber'
-        - $ref: '#/parameters/pageSize'
         - in: query
-          name: species
-          description: Search by pet species
+          name: name
```
ianbrown9475: For filter parameters, JSON:API convention is to use `filter[<name>]` where `<name>` usually corresponds to an attribute name. https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#filter-query-parameters. Most of our APIs don't follow this convention yet though.

openapi.yaml
```diff
@@ -25,105 +32,433 @@ security:
...
+          description: A list of pizzas
           schema:
-            $ref: '#/definitions/PetsResult'
-        '400':
-          $ref: '#/responses/400'
-        '500':
-          $ref: '#/responses/500'
+            $ref: "#/definitions/PizzasResults"
+        "500":
+          $ref: "#/responses/500"
```
ianbrown9475: What kind of response should the API return when one of the query parameters has a bad value? Like `?bakeTime=cat`?

openapi.yaml
```diff
@@ -25,105 +32,433 @@ security:
...
+        - pizza
       parameters:
         - in: body
-          name: body
+          name: attributes
+          description: attributes to assign to the new pizza
           required: true
           schema:
-            $ref: '#/definitions/PetPostBody'
+            $ref: "#/definitions/PizzaAttributes"
```
ianbrown9475: Could you change the request body to be a resource object here? https://jsonapi.org/format/#crud-creating

openapi.yaml
```diff
@@ -25,105 +32,433 @@ security:
...
+          description: The newly created pizza recipe
           schema:
-            $ref: '#/definitions/PetResult'
-        '400':
-          $ref: '#/responses/400'
-        '409':
-          description: Conflict
+            $ref: "#/definitions/PizzaResult"
+        "400":
+          $ref: "#/responses/400"
```
ianbrown9475: Another JSON:API convention is to respond with a 409 when `type` is invalid (or if a client-generated id is taken) https://jsonapi.org/format/#crud-creating-responses-409

openapi.yaml
```diff
@@ -25,105 +32,433 @@ security:
...
+            $ref: "#/definitions/PizzaResource"
+      responses:
+        "200":
+          description: The updated pizza recipe
+          schema:
+            $ref: "#/definitions/PizzaResult"
+        "400":
+          $ref: "#/responses/400"
+        "404":
+          $ref: "#/responses/404"
```
ianbrown9475: For PATCHing, a 409 is returned when type/id are invalid https://jsonapi.org/format/#crud-updating-responses-409

openapi.yaml
```diff
@@ -132,147 +467,202 @@ securityDefinitions:
...
-        description: Number of results per page
-        example: 25
-  SelfLink:
+      data:
+        type: array
+        items:
+          $ref: "#/definitions/PizzaResource"
+  PizzaResult:
     properties:
-      self:
```
ianbrown9475: Although optional, almost all of our APIs respond with a top-level self link. Could you add one here?

openapi.yaml
```diff
@@ -132,147 +467,202 @@ securityDefinitions:
...
+      data:
+        $ref: "#/definitions/PizzaResource"
+  PizzaResource:
+    properties:
+      type:
```
ianbrown9475: Most of our APIs have a resource-level self link if we support get-by-id requests to the resource. Could you add one here?

openapi.yaml
```diff
@@ -132,147 +467,202 @@ securityDefinitions:
...
+    type: object
+    properties:
+      name:
         type: string
-        format: url
-        description: The previous page of data
-      next:
+        description: the name of the ingredient
+        example: Sausage
+      type:
```
ianbrown9475: Looks like we can't have any attributes/relationships named "type" or "id": https://jsonapi.org/format/#document-resource-object-fields

openapi.yaml
jemisonf: Probably 400. I've added this in some of the other requests and should probably add it here as well.

openapi.yaml
jemisonf: interesting, can change this to ingredientType or something 👍

jemisonf:
> You could do the same thing for the ingredients resource. Or, you could make ingredients a property of the pizza resource. Whichever sounds better to you. Generally, when the resource/property in question has a lot of fields, it should be its own resource. What are your thoughts?

I think using resources is a good idea here from what I've seen, especially if it can simplify updating the pizza resource. I will look into adding that in this round of fixes/updates.

openapi.yaml
tsoliangwu0130: Actually `200` and `204` are both acceptable:

> If an update is successful and the server doesn’t update any attributes besides those provided, the server MUST return either a 200 OK status code and response document (as described above) or a 204 No Content status code with no response document.

We usually return a `200` if there's no sensitive data returned. Example: https://github.com/osu-mist/notes-api/blob/master/openapi.yaml#L167-L170
Just curious, why you choose `204` instead of `200` here?

jemisonf: 204 seemed simpler since there's no body to return. 200 makes sense if that's the standard for the team though.

jemisonf:
Updated a bunch of minor stuff based on feedback, and also added relationships. The implementation should be in line with the [JSON:API](https://jsonapi.org/format/#document-resource-object-relationships) spec for relationships.

openapi.yaml
```diff
@@ -132,147 +502,323 @@ securityDefinitions:
...
+                $ref: "#/definitions/SelfLink"
+              data:
+                type: array
+                description: a list of relationship resources with the ingredient type
+                items:
+                  $ref: "#/definitions/RelationshipResource"
+      included: &pizzaIncluded
+        type: array
+        items:
+          oneOf:
```
jemisonf: this creates an error in the swagger editor _but_ still produces the right output (a heterogeneous list of doughs and ingredients, see [here](https://jsonapi.org/format/#document-compound-documents))

openapi.yaml
```diff
@@ -25,105 +32,423 @@ security:
...
+          $ref: "#/responses/404"
+        "500":
+          $ref: "#/responses/500"
+    patch:
+      description: update the dough relationship for the pizza. Pass `null` as the data to remove.
+      tags:
+        - pizza
+        - dough
+      parameters:
+        - $ref: "#/parameters/pizzaId"
```
ianbrown9475: Could you add a parameter for the patch body?

openapi.yaml
```diff
@@ -25,105 +32,423 @@ security:
...
+          $ref: "#/responses/404"
+        "500":
+          $ref: "#/responses/500"
+    patch:
+      description: update *all* ingredients for a pizza.
+      tags:
+        - pizza
+        - ingredient
+      parameters:
+        - $ref: "#/parameters/pizzaId"
```
ianbrown9475: Similar comment, could you add a patch body here?

openapi.yaml
```diff
@@ -132,147 +457,273 @@ securityDefinitions:
...
+  RelationshipResourceList:
+    description: a list of relationship resources
+    properties:
+      data:
         type: array
         items:
-          $ref: '#/definitions/ErrorObject'
+          $ref: "#/definitions/RelationshipResource"
+parameters:
+  pizzaId:
```
ianbrown9475: Could you refactor the id parameters for doughs and ingredients in the same way as here?

openapi.yaml
```diff
@@ -132,147 +457,273 @@ securityDefinitions:
...
+        example: abc123
+      links:
+        $ref: "#/definitions/SelfLink"
+      attributes:
+        $ref: "#/definitions/PizzaAttributes"
+      relationships: &pizzaRelationships
+        properties:
+          dough:
+            properties:
+              data:
```
ianbrown9475: Could you also add a `links` object to this relationship?

jemisonf:
> * I noticed doughs and ingredients don't have a delete endpoint. Is this intentional?

Yes, this is to avoid issues with deleting an ingredient or dough and either "orphaning" pizzas that depends on that resource or having to delete all of the dependent pizzas.

> * What endpoints would I query to get a pizza's dough or a pizza's ingredients?

Those are:
Dough: `GET /pizzas/{pizzaId}/relationships/dough`
Ingredients: `GET /pizzas/{pizzaId}/relationships/ingredients`

openapi.yaml
```diff
@@ -25,105 +32,423 @@ security:
...
+      description: Update a pizza recipe
+      tags:
+        - pizza
+      parameters:
+        - $ref: "#/parameters/pizzaId"
+        - in: body
+          name: attributes
+          description: attributes of the pizza to be updated
+          schema:
+            $ref: "#/definitions/PizzaPostBody"
```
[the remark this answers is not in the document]
jemisonf: I will look into this. My impression at the time I wrote it was that the parameters were the same but I could be wrong about that.

openapi.yaml
```diff
@@ -132,147 +457,273 @@ securityDefinitions:
...
+
+  PizzaResource:
+    properties:
+      type: &pizzaType
         type: string
-        format: url
-        description: Self-link of current resource
-  PaginationLinks:
+        description: the type of the resource (pizza)
+        example: pizza
```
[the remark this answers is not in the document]
jemisonf: Good tip, thanks!

openapi.yaml
```diff
@@ -132,147 +457,273 @@ securityDefinitions:
...
+        type: array
+        items:
+          oneOf:
+            - $ref: "#/definitions/DoughRecipe"
+            - $ref: "#/definitions/IngredientResource"
+  PizzaPostBody:
     properties:
-      first:
+      type: *pizzaType
+      id: *pizzaId
```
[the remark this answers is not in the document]
jemisonf: Nope, good catch!

openapi.yaml
```diff
@@ -25,105 +32,423 @@ security:
...
+          description: attributes to assign to the new pizza
           required: true
           schema:
-            $ref: '#/definitions/PetPostBody'
+            $ref: "#/definitions/PizzaPostBody"
+      responses:
+        "201":
+          description: The newly created pizza recipe
+          schema:
+            $ref: "#/definitions/PizzaResult"
```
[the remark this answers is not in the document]
jemisonf: Added to other 201 responses as well

openapi.yaml
jemisonf: missed in this in the previous round of reviews; added to dough and ingredient response objects as well

openapi.yaml
```diff
@@ -132,147 +457,273 @@ securityDefinitions:
     scopes:
       full: Full access to the API
 definitions:
-  Meta:
+  PizzasResults:
```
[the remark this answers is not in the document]
jemisonf: Sounds good, I can switch to using this convention

openapi.yaml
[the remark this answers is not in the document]
jemisonf: I was wrong and there are some differences 👍

openapi.yaml
```diff
@@ -132,147 +462,375 @@ securityDefinitions:
...
+              links:
+                properties:
+                  self:
+                    type: string
+                    description: a link to the relationship
+                    example: https://api.oregonstate.edu/v1/pizzas/relationships/dough
+                  related:
+                    type: string
+                    description: a link to the dough referenced
+                    example: https://api.oregonstate.edu/v1/doughs/1
```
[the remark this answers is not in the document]
jemisonf: This was something I was kind of confused about when I was first reading this documentation -- this is talking about [related resource links](https://jsonapi.org/format/#document-resource-object-related-resource-links), right? Wondering if the spec could be more clear here 🤔

openapi.yaml
```diff
@@ -132,147 +462,375 @@ securityDefinitions:
...
+        required: [id, type]
+        properties:
+          type: *pizzaType
+          id: *pizzaId
+          attributes:
+            $ref: "#/definitions/PizzaPatchAttributes"
+          relationships: *pizzaUpdateRelationships
+  PizzaAttributes:
+    type: object
+    required: [name, ovenTemp, bakeTime]
```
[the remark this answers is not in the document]
jemisonf: Yeah, this makes sense. I'm going to change this so that `PizzaPatchAttributes` is now `PizzaAttributes`, which has no required fields, and so that `PizzaAttributes`, which has required fields, is now `PizzaPostAttributes`.

jemisonf:
> For the delete endpoints, you could always add support for deletions only when it wouldn't violate some other constraint such as if there's a pizza with a relationship to the ingredient/dough. What are your thoughts?

That would work, but would definitely add some extra development  overhead. It might not be very useful in the case that basically every ingredient was used by a pizza, which is what I assumed would happen. I also considered the option of making it possible to "archive" ingredients (or some similar concept) so that they couldn't be added to new pizzas but wouldn't disappear from the database. My initial thought though was that this functionality was out of scope of the design at this phase in the process.

openapi.yaml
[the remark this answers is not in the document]
jemisonf: Related to your comment here:
> but what if a client needed to fetch the actual resources?

Going to add a related resource link for dough and ingredients as a solution

openapi.yaml
```diff
@@ -25,105 +32,466 @@ security:
...
+        - in: query
+          name: filter[name]
+          description: Search ingredients by name
+          required: false
+          type: string
+        - in: query
+          name: filter[ingredientType]
+          description: Search ingredients by type
+          required: false
+          type: string
```
tsoliangwu0130: Should this be `enum`?

openapi.yaml
```diff
@@ -132,147 +500,417 @@ securityDefinitions:
...
+              data:
+                type: array
+                description: a list of relationship resources with the ingredient type
+                items:
+                  $ref: "#/definitions/IngredientRelationshipResource"
+      included: &pizzaIncluded
+        type: array
+        items:
+          type: object
+          description: An ingredient resource or a dough resource
```
tsoliangwu0130: Can't ingredients be included?

jemisonf:
Takeaways from review meeting:
* Add `relationships` key to POST/PATCH body for pizzas
* Add `DELETE` endpoint for pizza->ingredients relationship
* Add notes to ingredients
* Investigate adding additional properties to pizza->ingredients relationship

openapi.yaml
jemisonf: Yeah, this list can include both ingredient and dough resources, but there's no way to represent that in the swagger spec at the moment

jemisonf:
> Investigate adding additional properties to pizza->ingredients relationship

This can be done through the `meta` field of the relationship object. Possibly a non-standard usage of meta, but I will try it out and see how it looks.

jemisonf:
> Find correct error codes for relationships 

Going with 403 for this

openapi.yaml
```diff
@@ -850,11 +886,18 @@ definitions:
         type: string
         description: the unique ID of the resource
         example: abc123
+      meta:
+        properties:
+          description:
+            type: string
+            description: information about the way the ingredient is used in the pizza
+            example: only put pepperoni on half of the pizza
+
```
jemisonf: This is the only way I can find to embed information like quantity of an ingredient in the ingredient relationship itself.

jemisonf:
An FYI on the question about `POST`ing the pepperoni relationship again when it already exists:

> If a client makes a POST request to a URL from a relationship link, the server MUST add the specified members to the relationship unless they are already present. If a given type and id is already in the relationship, the server MUST NOT add it again.

From [JSON:API docs](https://jsonapi.org/format/#crud-updating-to-many-relationships)

openapi.yaml
```diff
@@ -132,120 +556,313 @@ securityDefinitions:
     scopes:
       full: Full access to the API
 definitions:
-  Meta:
+  PizzaSetResults:
```
tsoliangwu0130: PizzaSetResults -> PizzaSetResult

openapi.yaml
```diff
@@ -25,105 +32,522 @@ security:
...
+          description: Successfully removed the ingredients
+        '400':
+          $ref: '#/responses/400'
+        '403':
+          $ref: '#/responses/403'
+        '404':
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+  /pizzas/{pizzaId}/ingredients:
```
tsoliangwu0130: Hmm, I'm not really sure, but should we add `include` parameter to `/pizzas/{pizzaID}`? If so, won't the result from `/pizzas/{pizzaId}?include=ingredients` overlap the result with this endpoint? Since the current endpoint is only for `GET` method now, unless you want to provide more filters here, I guess I'm not really sure what's the main purpose of to have a new endpoint (and also `/pizzas/{pizzaId}/dough`)? Or maybe this's mainly for the related link?

@jemisonf what's your thought on it?
@ianbrown9475 can you also provide some feedback related to this since you've worked on `relationship` stuff before?

openapi.yaml
```diff
@@ -25,105 +32,522 @@ security:
...
+            $ref: '#/definitions/IngredientRelationshipResourceList'
+        '400':
+          $ref: '#/responses/400'
+        '403':
+          $ref: '#/responses/403'
+        '404':
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+    delete:
```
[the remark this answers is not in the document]
Ruefa: I also don't remember seeing this in the meeting yesterday. I think the description is a bit ambiguous. Does this delete all ingredients from the pizza or just the ones passed in the body?

openapi.yaml
Ruefa: I agree this seems redundant. Less endpoints will be easier to maintain.

openapi.yaml
```diff
@@ -25,105 +32,522 @@ security:
...
+          $ref: '#/responses/400'
+        '404':
+          $ref: '#/responses/404'
+        '409':
+          $ref: '#/responses/409'
+        '500':
+          $ref: '#/responses/500'
+  /ingredients:
+    get:
+      description: Get the a list of all ingredients
```
Ruefa: Grammar error here

openapi.yaml
```diff
@@ -132,120 +556,313 @@ securityDefinitions:
...
-      prev:
+        description: the name of the ingredient
+        example: Sausage
+      ingredientType:
         type: string
-        format: url
-        description: The previous page of data
-      next:
+        description: the type of the ingredient
+        enum: [meat, vegetable, fruit, other]
```
Ruefa: Been a bit since I wrote openapi. I see this enum list both here and above on one of the endpoints. Is there a way to define the enum list and reference it when needed?

openapi.yaml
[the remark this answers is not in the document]
jemisonf: It just deletes the ones in the body. You can delete all by `PATCH`ing a blank data section.

openapi.yaml
```diff
@@ -25,105 +32,522 @@ security:
...
+          name: filter[ovenTemp]
+          description: Search pizzas by bake temperature (in degrees Fahrenheit)
           required: false
+          type: integer
+        - in: query
+          name: filter[bakeTime]
+          description: Search pizzas by baking time (in minutes)
+          type: integer
+        - in: query
+          name: filter[ingredientIds]
```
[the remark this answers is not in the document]
jemisonf: I can add both of those

openapi.yaml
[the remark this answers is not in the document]
jemisonf: Details in JSON:API spec here: https://jsonapi.org/format/#crud-updating-to-many-relationships

openapi.yaml
jemisonf: I don't think you can do it with openAPI but you could use yaml anchors/aliases.

jemisonf:
Discussion with @tsoliangwu0130: the [mist style guide](https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#filter-query-parameters) has no guidelines for `filter` parameters to restrict results to resources containing none of the specified values, i.e. searching for all pizzas that do not have pepperoni and pineapple as an ingredient. 

We decided to use a `noneOf` parameter for this API, and felt it would be a good idea to add that to the style guide as well.

There's also an inconsistency in the formatting of the parameter names; some are postfixed with -Of, like `oneOf`, and some are not, like `all` or `some`. We thought it would make sense to have all of the parameters use the same style, something like:
* oneOf
* some => someOf
* all => allOf
* + oneOf

@ianbrown9475 do you have any thoughts on this since you wrote the original guidelines?

tsoliangwu0130:
Since Swagger/OpenAPI 3 use `anyOf`, should we also use the same `keyword` instead of `someOf`?

openapi.yaml
jemisonf: > Or maybe this's mainly for the related link

This is the reason at this point. If the related link isn't important then it makes sense to take both endpoints out.

jemisonf: Here's the usage for those:

In DoughRelationshipResourceSingleton
```yaml
      links:
        properties:
          self:
            type: string
            example: https://api.oregonstate.edu/v1/pizzas/1/relationships/dough
          related:
            type: string
            example: https://api.oregonstate.edu/v1/pizzas/1/dough
```

in IngredientRelationshipResourceList
```yaml
      links:
        properties:
          self:
            type: string
            example: https://api.oregonstate.edu/v1/pizzas/1/relationships/ingredients
          related:
            type: string
            example: https://api.oregonstate.edu/v1/pizzas/1/ingredients
```

tsoliangwu0130: That's fair. I'm fine with keeping both for now. Just a little not sure should we maintain two endpoints for almost the same function just because of the related link.

@ianbrown9475  what's your thought on it?

openapi.yaml
```diff
@@ -132,120 +580,313 @@ securityDefinitions:
...
+    properties:
+      data:
+        required: [id, type]
+        properties:
+          type: *pizzaType
+          id: *pizzaId
+          attributes:
+            $ref: '#/definitions/PizzaAttributes'
+          relationships: *pizzaUpdateRelationships
+  PizzaPostAttributes:
```
tsoliangwu0130: Looks like this definition is defined but never used. Should this be removed?

openapi.yaml
```diff
@@ -733,7 +733,7 @@ definitions:
       ingredientType:
         type: string
         description: the type of the ingredient
-        enum: [meat, vegetable, fruit, other]
+        enum: [cheese, sauce, meat, vegetable, fruit, other]
```
tsoliangwu0130: This enum list was repeated with the parameter `filter[ingredientType]` at `/ingredients` endpoint. Can you reuse the enum there?

jemisonf: It looks like this is an [openapi 3.0 feature](https://swagger.io/docs/specification/data-models/enums/) but I'm not sure there's a good way to do it here

tsoliangwu0130: Take a look at this [example](https://github.com/osu-mist/parks-api/blob/cc322d795d245380bc7cc71ea6ebaf96a9a3a8c6/openapi.yaml#L264). Let me know if that still doesn't work.

openapi.yaml
jemisonf: This should be used on line 659 instead of `"#/definitions/PizzaAttributes"`, in the definition of PizzaPostBody. I'll add it there

openapi.yaml
jemisonf: That works. My only hesitancy with that approach was having to define the anchor in a place where it doesn't feel like it would "naturally live", ie at the top of the document in the pizza filters. But you're probably right that it's worth it to not repeat the definition.

ianbrown9475:
> @ianbrown9475 do you have any thoughts on this since you wrote the original guidelines?

I think the original intent was to differentiate the operators that compare a single value to an array of filter parameters and those that compare an array of values to an array of filter parameters.

If I'm understanding this right, you were wanting to add `noneOf` to check if a pizza has none of the following ingredients, right? Meaning a pizza might have ingredients `[pepperoni, olives]` and a query might be `?ingredients[noneOf]=pineapple,spinach`. Will `noneOf` also be used for single-value fields and array-value queries? Like a pizza might have `ovenTemp: 500` and a query might be `?ovenTemp[noneOf]=400,425,450`.

I think we should try and have different names for these since the behavior is slightly different. I think one of the original thoughts for array values with array queries was `hasAll`, `hasSome`, and maybe `hasNone` for this case. What do you think?

openapi.yaml
ianbrown9475: I think if we choose to just have one of these types of endpoints, we should use `/resources/{id}/relatedResource` instead of just having `/resources/{id}?include=relatedResource`. The compound documents seem to be more of a convenient feature for reducing requests where necessary. Having related resources be accessible seems more like basic functionality. For example, if we represented the students API using relationships like this: https://github.com/osu-mist/edu-students-api/blob/master/openapi.yaml, we wouldn't want to remove all of the `/students/{id}/subResource` endpoints, even if we supported compound documents. It seems like we would be overloading that query parameter and making the related resources less discoverable. There could also be cases where the main resource contains a lot of information and if a client only wants related resources, compound documents would add too much unnecessary information. @tsoliangwu0130 @jemisonf what are your thoughts on this?

openapi.yaml
```diff
@@ -132,120 +586,313 @@ securityDefinitions:
...
+              dough:
+                properties:
+                  data:
+                    $ref: '#/definitions/DoughRelationshipResource'
+              ingredients:
+                properties:
+                  data:
+                    type: array
+                    items:
+                      $ref: '#/definitions/IngredientRelationshipResource'
```
ianbrown9475: `properties` and `ingredients` definitely make sense to make optional, but I think some of the fields within these objects should be required. Could you take a look at this? I think this will need to be fixed for the other endpoints as well.

openapi.yaml
```diff
@@ -25,105 +32,552 @@ security:
...
+      description: update the dough relationship for the pizza. Pass `null` as the data to remove.
+      tags:
+        - pizza
+        - dough
+      parameters:
+        - $ref: '#/parameters/pizzaId'
+        - in: body
+          name: dough
+          schema:
+            $ref: '#/definitions/DoughRelationshipResourceSingleton'
```
ianbrown9475: In this request body, I think `data` needs to be required. Could you also fix this for the other PATCH relationships endpoints?

jemisonf:
> I think the original intent was to differentiate the operators that compare a single value to an array of filter parameters and those that compare an array of values to an array of filter parameters.

I think this is actually a really good idea. I'm not sure if the current solution does that unambiguously, but I'm also not sure it would be possible to perfectly accomplish that goal. One thought I had would be to use `in` instead of `oneOf`, but that might only clear things up for python programmers. Then we could also have `notIn` for single value fields, and `none` for multi-value fields. Thoughts?

openapi.yaml
jemisonf: I think that's a good point. Another reason to not use compound documents is that it would simplify implementation (I think), and we could always add them later (in theory) without breaking the API for users. It's a hard tradeoff to evaluate when you don't have any real users :/

openapi.yaml
jemisonf: Adding the `type` and `id` fields as required on each relationship resource object

openapi.yaml
tsoliangwu0130: @ianbrown9475 That sounds fair. I guess we need to think about this question a little more in case we run into similar issues in the future.

1. When should we support the compound endpoint (`/resources/{id}?include=relatedResource`?
2. Should compound endpoint only be used as a convenient feature?
3. Is there any scenario that we only want the compound endpoint but not subresource endpoint (`/resources/{id}/relatedResource`) ?

**Compound endpoint**

  Pros:
  1. Convenient and could reduce requests.

  Cons:
  1. Could possiblly return too many unnecessary information and make the return data too large and hard to read. And it can potentially slow down the API performance.
  2. We may need to maintain the same/similar logic twice for both compond and subresource endpoint.
  3. Can portentially increase the complexity of implementing/maintaining unit tests. 
  4. Increase the diffculty of protecting data. For example, if a client is not allowed to access the subresource endpoint, what should we do if he/she try to include the subresource at the base endpoint request?

**Subresource endpoint**

  Pros:
  1. Basic functionality and may support other HTTP methods like `PATCH` and `DELETE`.
  2. Provide related link.

  Cons:
  1. Clients are fourced to make at least two requests to get basic data + subresource data for an resource. It can be much worse if we have more than one subresource endpoints. Think about the new DX project from WAMS team, on the dashbaord page, they'll need to visit all subresources endpoint which means they will need to make about 11 request for each student.

@jemisonf @Ruefa @kalcicd @ianbrown9475 let me know if I miss anything and feel free to add your thoughts on it.

jemisonf: That looks like a really good pro/cons list to me. I'm wondering if this project would be a good chance to experiment with using compound endpoints and get a clearer sense of how much work it actually is to implement. Being able to avoid this problem:

> Clients are forced to make at least two requests to get basic data + subresource data for a resource. It can be much worse if we have more than one subresource endpoints. Think about the new DX project from WAMS team, on the dashboard page, they'll need to visit all subresources endpoint which means they will need to make about 11 requests for each student.

Is a big deal and I think is worth investigating.

ianbrown9475:
@jemisonf I think the only issue with `in`/`notIn` is that it seems like it could be applied to strings (like `'a' in 'abc' == True` in python). Maybe we can discuss this with the rest of the group to get other opinions.

openapi.yaml
tsoliangwu0130: @jemisonf great idea. Let's use this learning project as a POC for the `relationships`feature and implement both kinds of endpoints and see how things going.

Feel free to document whatever you found in this [CO-1479](https://jira.sig.oregonstate.edu/browse/CO-1479). Thanks for doing that 😄

PR #3, opened 2019-10-08
CO-1598: Adds DB setup file

jemisonf:
Initially based on code auto-generated by SQL Developer. Also adds a sample pizza with a dough and 3 ingredients.


src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,72 @@
...
+	 CONSTRAINT "PIZZAS_PK" PRIMARY KEY ("ID") ENABLE,
+	 CONSTRAINT "PIZZAS_FK1" FOREIGN KEY ("DOUGH_ID")
+	  REFERENCES "DOUGHS" ("ID") ON DELETE SET NULL ENABLE
+   );
+
+CREATE TABLE "PIZZA_INGREDIENTS" 
+   (	"PIZZA_ID" NUMBER NOT NULL ENABLE, 
+	"INGREDIENT_ID" NUMBER NOT NULL ENABLE, 
+	"DESCRIPTION" VARCHAR2(512 BYTE), 
+	 CONSTRAINT "PIZZA_INGREDIENTS_UK1" UNIQUE ("PIZZA_ID", "INGREDIENT_ID") ENABLE,
```
jemisonf: Not having a primary key but requiring the combination of pizza id and ingredient id jumped out to me as the most obvious solution for this table. Not sure it's the best.

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,73 @@
...
+	"FLOUR_TYPE" VARCHAR2(256 BYTE) NOT NULL ENABLE, 
+	"WATER_TEMP" NUMBER(*,0) NOT NULL ENABLE, 
+	"GRAMS_YEAST" NUMBER(*,0) NOT NULL ENABLE, 
+	"GRAMS_SALT" NUMBER(*,0) NOT NULL ENABLE, 
+	"GRAMS_SUGAR" NUMBER(*,0) NOT NULL ENABLE, 
+	"GRAMS_OLIVE_OIL" NUMBER(*,0) NOT NULL ENABLE, 
+	"BULK_FERMENT_TIME" NUMBER(*,0) NOT NULL ENABLE, 
+	"PROOF_TIME" NUMBER(*,0) NOT NULL ENABLE, 
+	"SPECIAL_INSTRUCTIONS" VARCHAR2(512 BYTE), 
+	 CONSTRAINT "DOUGHS_PK" PRIMARY KEY ("ID") ENABLE
```
jemisonf: It seems like there's a bunch of ways to define constraints in oracle sql. Not sure if there's a preferred method.

jemisonf:
For context: this probably shouldn't be merged until #2 is done but we have the structure of the data pretty much totally worked out so I'm working on this to avoid being blocked on #2.

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,72 @@
+DROP TABLE PIZZA_INGREDIENTS;
+DROP TABLE PIZZAS;
+DROP TABLE INGREDIENTS;
+DROP TABLE DOUGHS;
+
+CREATE TABLE "DOUGHS" 
+   (	"ID" NUMBER GENERATED ALWAYS AS IDENTITY MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 1 START WITH 1 CACHE 20 NOORDER  NOCYCLE  NOT NULL, 
+	"NAME" VARCHAR2(256 BYTE) NOT NULL, 
+	"GRAMS_FLOUR" NUMBER(*,0) NOT NULL,
```
jemisonf: this is generated by sql developer to indicate an integer. Arbitrary precision seems fine to me but there is a relatively low theoretical cap for all of these value so we could specify a smaller precision if we wanted to go there.

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,72 @@
+DROP TABLE PIZZA_INGREDIENTS;
+DROP TABLE PIZZAS;
+DROP TABLE INGREDIENTS;
+DROP TABLE DOUGHS;
+
+CREATE TABLE "DOUGHS" 
+   (	"ID" NUMBER GENERATED ALWAYS AS IDENTITY MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 1 START WITH 1 CACHE 20 NOORDER  NOCYCLE  NOT NULL, 
+	"NAME" VARCHAR2(256 BYTE) NOT NULL,
```
jemisonf: I've used 256 bytes for "short" strings, like names, and 512 bytes for "long" strings like descriptions.

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,72 @@
...
+	"PROOF_TIME" NUMBER(*,0) NOT NULL, 
+	"SPECIAL_INSTRUCTIONS" VARCHAR2(512 BYTE), 
+	 PRIMARY KEY ("ID")
+   );
+
+CREATE TABLE "INGREDIENTS" 
+   (	"TYPE" VARCHAR2(20 BYTE) NOT NULL, 
+	"NAME" VARCHAR2(256 BYTE) NOT NULL, 
+	"NOTES" VARCHAR2(256 BYTE), 
+	"ID" NUMBER GENERATED ALWAYS AS IDENTITY MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 1 START WITH 1 CACHE 20 NOORDER  NOCYCLE  NOT NULL,
```
Ruefa: You don't need to set a min and max value for ID https://github.com/osu-mist/game-curator-api/blob/4929a92be94da94b9fb2084382360c9832da033b/api/v1/db/oracledb/setupDB.sql#L6

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -0,0 +1,72 @@
...
+    values ('Sample dough', 500, 400, 'Bread flour', 90, 10, 10, 0, 50, 120, 30, 'Can also proof overnight in the refrigerator');
+    
+INSERT INTO INGREDIENTS (NAME, TYPE, NOTES)
+    values ('Pepperoni', 'meat', NULL);
+INSERT INTO INGREDIENTS (NAME, TYPE, NOTES)
+    values ('Low moisture mozarella', 'cheese', NULL);
+INSERT INTO INGREDIENTS (NAME, TYPE, NOTES)
+    values ('Marinara sauce', 'sauce', 'Store bought pizza or marinara sauce is fine');
+    
+INSERT INTO PIZZAS (NAME, BAKE_TIME, OVEN_TEMP, SPECIAL_INSTRUCTIONS, DOUGH_ID)
```
Ruefa: You're going to want at least 3 pizzas and doughs.

src/api/v1/db/oracledb/setupDb.sql
jemisonf: According to [this](https://www.oracletutorial.com/oracle-basics/oracle-number-data-type/) it has the maximum range and precision by default

src/api/v1/db/oracledb/setupDb.sql
jemisonf: Thanks, it looks like the cache stuff is optional as well

PR #5, opened 2019-10-23
CO-1620: adds doughs unit tests

jemisonf:


jemisonf:
(this shouldn't be merged until after #4)

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import sinon from 'sinon';
+
+import { getDoughs } from 'api/v1/db/oracledb/doughs-dao';
+import * as doughsSerializer from 'api/v1/serializers/doughs-serializer';
+import * as connectionApi from 'api/v1/db/oracledb/connection';
+// import { connection } from 'oracledb';
```
tsoliangwu0130: Can you remove this comment?

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import sinon from 'sinon';
+
+import { getDoughs } from 'api/v1/db/oracledb/doughs-dao';
+import * as doughsSerializer from 'api/v1/serializers/doughs-serializer';
+import * as connectionApi from 'api/v1/db/oracledb/connection';
```
tsoliangwu0130: This should be imported before `import { getDoughs } from 'api/v1/db/oracledb/doughs-dao';`

> Each line within groups 1-3 should be ordered alphabetically by the first alphanumeric character in the name of the package/file in the statement.

Ref: https://github.com/osu-mist/style-guides/blob/master/JavaScript.md#importing-modules

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+describe('test doughs dao', () => {
+  let serializeDoughsStub = null;
+  let getConnectionStub = null;
+  let connectionSpy = null;
+  const getDoughsConditions = 'ID AS "id", NAME AS "name", GRAMS_FLOUR AS "gramsFlour", GRAMS_WATER AS "gramsWater", FLOUR_TYPE AS "flourType", WATER_TEMP AS "waterTemp", GRAMS_YEAST AS "gramsYeast", GRAMS_SALT AS "gramsSalt", GRAMS_SUGAR AS "gramsSugar", GRAMS_OLIVE_OIL AS "gramsOliveOil", BULK_FERMENT_TIME AS "bulkFermentTime", PROOF_TIME AS "proofTime", SPECIAL_INSTRUCTIONS AS "specialInstructions"';
+  const getDoughsQuery = `SELECT ${getDoughsConditions} FROM DOUGHS `;
+  const getDoughsQueryWithWaterTemp = `SELECT ${getDoughsConditions} FROM DOUGHS WHERE WATER_TEMP = :waterTemp ;`;
+  const getDoughsQueryWithMultipleConditions = `SELECT ${getDoughsConditions} FROM DOUGHS WHERE WATER_TEMP = :waterTemp AND GRAMS_WATER = :gramsWater AND PROOF_TIME = :proofTime AND NAME = :name ;`;
+  const executeReturnRows = ['a', 'b', 'c'];
+  const executeReturn = { rows: executeReturnRows, metadata: 'ignore-this' };
```
tsoliangwu0130: All of these test data should be collected in a separate file like [here](https://github.com/osu-mist/students-api/blob/master/src/tests/unit/test-data.js). It'll make your code easier to read and maintain.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+    connectionSpy = sinon.stub().returns(executeReturn);
+    serializeDoughsStub = sinon.stub(doughsSerializer, 'serializeDoughs');
+    getConnectionStub = sinon.stub(connectionApi, 'getConnection');
+    getConnectionStub.resolves({
+      execute: connectionSpy,
+      close: () => { },
+      commit: () => { },
+    });
+    serializeDoughsStub.returns(baseGetDoughsReturn);
+  });
```
tsoliangwu0130: Clever, but you still need to connect to the real DB and stub it right? When running unit tests, we shouldn't touch the real DB instance in case we screw things up. In other words, the unit test shouldn't rely on the config file at all. The unit tests won't work if you run without a config file.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+  describe('getDoughs', () => {
+    it('returns the output from serializeDoughs', () => {
+      const result = getDoughs({});
+      return result.should.eventually.deep.equal(baseGetDoughsReturn);
+    });
+    it('extracts only the rows from the database return', async () => {
+      await getDoughs({});
```
tsoliangwu0130: You don't need to `await` here. Instead, you should use the keyword `eventually` from [`chai-as-promised`](https://www.chaijs.com/plugins/chai-as-promised/) to asset the behavior of promises.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+          'filter[gramsWater]': 200,
+          'filter[proofTime]': 90,
+          'filter[name]': 'Sample dough',
+        });
+        connectionSpy.getCall(0).calledWith(getDoughsQueryWithMultipleConditions, {
+          waterTemp: 90,
+          gramsWater: 200,
+          proofTime: 90,
+          name: 'Sample dough',
+        });
```
tsoliangwu0130: Your test should also cover every function (include helpers like `doughColumnAliases`, `normalizeFilterName`, etc.).

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+chai.use(chaiAsPromised);
+
+const baseGetDoughsReturn = {
+  data: [],
+};
+
+describe('test doughs dao', () => {
+  let serializeDoughsStub = null;
+  let getConnectionStub = null;
+  let connectionSpy = null;
```
tsoliangwu0130: I don't think you need to give a initial `null` to them.

```js
> let a;
undefined
> a = 'A'
'A'
```

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+    sinon.restore();
+  });
+  describe('getDoughs', () => {
+    it('returns the output from serializeDoughs', () => {
+      const result = getDoughs({});
+      return result.should.eventually.deep.equal(baseGetDoughsReturn);
+    });
+    it('extracts only the rows from the database return', async () => {
+      await getDoughs({});
+      serializeDoughsStub.getCall(0).calledWith(executeReturnRows).should.equal(true);
```
tsoliangwu0130: What does `true` mean? You should use `deepEqual` to compare the actual result. Example: https://github.com/osu-mist/terms-api/blob/b9acb21bb42fa5a483af15f514ea9017f4366c9f/src/tests/unit/terms-dao.test.js#L72-L74

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+    describe('when it has an invalid filter', () => {
+      it('does not parse the invalid filter', async () => {
+        await getDoughs({ abc: 'def' });
+        connectionSpy.getCall(0).calledWith(getDoughsQuery, {}).should.equal(true);
+      });
+    });
+    describe('when it has valid filters', () => {
+      it('properly parses those filters into bind paramters', async () => {
+        await getDoughs({ 'filter[waterTemp]': 90 });
+        connectionSpy.getCall(0).calledWith(getDoughsQueryWithWaterTemp, { waterTemp: 90 });
```
tsoliangwu0130: What's the expected result of this assertion?

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+      });
+    });
+    describe('when it gets valid and invalid filters', () => {
+      it('only parses the valid filters into bind parameters', async () => {
+        await getDoughs({ 'filter[waterTemp]': 90, abc: 'def' });
+        connectionSpy.getCall(0).calledWith(getDoughsQueryWithWaterTemp, { waterTemp: 90 });
```
tsoliangwu0130: What's the expected result of this assertion?

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+    describe('when it gets multiple valid filters', () => {
+      it('properly parses all of those filters', async () => {
+        await getDoughs({
+          'filter[waterTemp]': 90,
+          'filter[gramsWater]': 200,
+          'filter[proofTime]': 90,
+          'filter[name]': 'Sample dough',
+        });
+        connectionSpy.getCall(0).calledWith(getDoughsQueryWithMultipleConditions, {
```
tsoliangwu0130: What's the expected result of this assertion?

tsoliangwu0130:
I also ran into the following errors, can you take a look at them?

```
id filter
           does not parse the invalid filter:
     Error: Request includes invalid filters
      at processGetFilters (dist/api/v1/db/oracledb/api/v1/db/oracledb/doughs-dao.js:95:11)
      at processGetFilters (dist/api/v1/db/oracledb/api/v1/db/oracledb/doughs-dao.js:113:40)
      at Context.it (dist/tests/unit/tests/unit/doughs-dao.test.js:52:15)

  2) test doughs dao
       getDoughs
         when it gets valid and invalid filters
           only parses the valid filters into bind parameters:
     Error: Request includes invalid filters
      at processGetFilters (dist/api/v1/db/oracledb/api/v1/db/oracledb/doughs-dao.js:95:11)
      at processGetFilters (dist/api/v1/db/oracledb/api/v1/db/oracledb/doughs-dao.js:113:40)
      at Context.it (dist/tests/unit/tests/unit/doughs-dao.test.js:64:15)

```

src/tests/unit/doughs-dao.test.js
jemisonf: The DB should be stubbed here:
```js
getConnectionStub = sinon.stub(connectionApi, 'getConnection');
    getConnectionStub.resolves({
      execute: connectionSpy,
      close: () => { },
      commit: () => { },
    });
```
which stops the `getConnection` method from returning a real database connection

src/tests/unit/doughs-dao.test.js
jemisonf: The problem there is that I'm trying to assert on a side effect of the function call, ie how it interacts with the connection, rather than on the return of the function. I will keep looking into this but I didn't see a way to do that with the chai as promised matchers.

src/tests/unit/doughs-dao.test.js
jemisonf: My thought there was that those are implicitly covered in tests that assume the bind parameters will be correctly parsed, and that covering them explicitly would mean testing unexported functions that aren't part of the interface. I think that if were were to look at code coverage for these tests they would in fact be hitting all of the helper functions in the file.

src/tests/unit/doughs-dao.test.js
jemisonf: In this case it's testing that `calledWith` returns the literal value `true`.  I think that I can use the `.be.true` matcher to check that more explicitly.

src/tests/unit/doughs-dao.test.js
jemisonf: Haha, no wonder these tests were all passing . . .

jemisonf:
> I also ran into the following errors, can you take a look at them?

@tsoliangwu0130 for this make sure you've rebased the latest master branch. I think this is triggered by the code change that got accidentally merged with #4 .

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Yeah, I see what you mean. My main concern is you still imported the connection [at the very beginning here](https://github.com/osu-mist/pizza-api/blob/d29ca1a3f2c89af5a4f1f6d25a9c79cfb8e4b42e/src/tests/unit/doughs-dao.test.js#L7) with the real config data.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Hmm, not sure if I understand what you mean, but are you trying to assert if `serializeDoughsStub` is `calledWith` `executeReturnRows`? If so, you don't even need `should.equal(true)` or `should.be.true` since `calledWith` is already an assertion: https://www.chaijs.com/plugins/sinon-chai/

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: No really. The main point of writing unit tests is to make sure every function should work as expected and be tested. Some corner cases could be easily ignored if you only implicitly test them by calling parents functions/methods.

tsoliangwu0130:
@jemisonf Hmm, why do I need to rebase the master? I only ran test on this branch, why will it matter if I rebase master or not?

src/tests/unit/doughs-dao.test.js
jemisonf: The only side effects from the import are these lines though from the start of the file:
```js
const dbConfig = config.get('dataSources').oracledb;

process.on('SIGINT', () => process.exit());
oracledb.outFormat = oracledb.OBJECT;
oracledb.fetchAsString = [oracledb.DATE, oracledb.NUMBER];

/** Increase 1 extra thread for every 5 pools but no more than 128 */
const threadPoolSize = dbConfig.poolMax + (dbConfig.poolMax / 5);
process.env.UV_THREADPOOL_SIZE = threadPoolSize > 128 ? 128 : threadPoolSize;

/** Connection pool */
let pool;
```

And I don't see anything that would actually trigger it to connect to the database. I think you'd have to call the actual `getConnection` or `getPool` functions to do that. I'm also not totally sure how to mock `getConnection` without using this import statement. I've seen `proxyquire` used in other API code but since I'm using the es6 module import syntax I'm not sure if that will work here.

jemisonf:
> Hmm, why do I need to rebase the master? I only ran test on this branch, why will it matter if I rebase master or not?

@tsoliangwu0130 Nope, I misspoke on that. I needed to rebase master to this branch myself and then push the changes, which I've now done. If you pull the most up-to-date version of the branch it should be fun.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Yes, using `proxyquire` should be the correct way. Here're some examples:

Terms API: https://github.com/osu-mist/terms-api/blob/b9acb21bb42fa5a483af15f514ea9017f4366c9f/src/tests/unit/terms-dao.test.js#L15-L35
OnBase API: https://github.com/osu-mist/onbase-api/blob/cd088d37d74986168a2a4e6cafc2db4f4ec58e19/src/tests/unit/onbase-dao.test.js#L26-L44
Students API: https://github.com/osu-mist/students-api/blob/2ad5a02bb99043bc7ae8ca2571fd836ea2a8a300/src/tests/unit/students-dao.test.js#L18-L32

src/tests/unit/doughs-dao.test.js
jemisonf: Oh very cool, I'll definitely use the `should.have.been` matcher instead

tsoliangwu0130: Cool 👍
Same with other comments I left, you probably also wanna assert what's the expected result of it.

jemisonf: I think using `should.have.been.calledWith(...)` is the same as `calledWith(...).should.be.true` or `calledWith(...).should.deepEqual(true)` based on reading those docs

tsoliangwu0130: Oh, I think I know what's the confusing part here lol.
Yeah it should be totally fine to use either `should.have.been.calledWith(...)` or `calledWith(...).should.be.true` to assert the spy object was called correctly. What I mean for `deepEqual` is we also need to asset the return result from `getDoughs(getDoughsData.mixedValidFilters)` is deeply equal to something.

src/tests/unit/doughs-dao.test.js
jemisonf: (Sorry for the long comment)

My assumption was always that a "unit" in this case is a single module, and we're testing the interface of the unit by testing the exported functions. `doughColumnAliases` and`normalizeFilterName` are internal to the `doughs-dao` module and subject to change, so testing their specific behavior seems like it would make our tests more brittle. They're also not all that complicated--`doughColumnAliases` in particular is constant and generates the same value every time. 

The behavior we care about, in my opinion, is how the `getDoughs` function generates queries and bind parameters, and how it interacts with `serializeDoughs`. That functionality fully encompasses properly normalizing filter names, and if we introduce a bug into the filter normalization functions it should propagate into our existing tests.

I do understand the alternative position though, and at the end of the day I'm not going to die on this hill. This discussion also kind of makes me wonder if it would make sense to pull some of this logic out into a separate util module that we could then test on its own, without changing the interface of `doughs-dao`.

src/tests/unit/doughs-dao.test.js
jemisonf: Ah, my original thought was that this test was only to check the values the spy was called with, and that checking the return value was handled in other tests.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: I'm happy to discuss this topic deeper 😀 

Actually this is a debatable and interesting topic. You can find many threads about it like [this](https://stackoverflow.com/questions/1861338/unit-test-helper-methods) and [this](https://dzone.com/articles/unit-testing-guidelines-what-to-test-and-what-not).

Take a look at this post: http://softwaretestingfundamentals.com/unit-testing/

> Do not create test cases for everything. Instead, focus on the tests that impact the behavior of the system.

but also

> Write test cases that are independent of each other. For example, if a class depends on a database, do not write a case that interacts with the database to test the class. Instead, create an abstract interface around that database connection and implement that interface with a mock object.

To me, sometimes these principles can violate with each other. The question here is how "small" unit we wanna test. Let me rephrase my statement. I think what I meant to test "everything function" is not really write test cases for every single function, instead, we should test every function which directly or indirectly impact the result. For example, if you have a helper function which only takes care of generating logging messages or error messages and the logic is pretty simple, then I totally agree it's not necessary to test it. However, in your case here since most of these helper functions are used to generate SQL query and will be executed against the DB, I guess I lean toward to test them to make sure it won't generate any bad query (like SQL injection).

I'm open to any thoughts, feel free to add your comment or concern here as well 😄 

@Ruefa what's your thought on it?

jemisonf: > However, in your case here since most of these helper functions are used to generate SQL query and will be executed against the DB, I guess I lean toward to test them to make sure it won't generate any bad query (like SQL injection).

I do agree it's really important to test SQL generation, but we do have a way of doing that by testing the values that `connection.execute()` is called with. The implementation details of how that SQL is generated aren't as important, in my opinion.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Oh, did I miss anything? Which test case tested the return value?

jemisonf: The test `It('returns the output from serializeDoughs')` checks that, since the return value comes directly from `serializeDoughs` without any other processing.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Hmm, not sure.
To assert some functions will call `connection.execute()` with an expected query and the SQL query generator will always guarantee to generate the correct query are still slightly different to me.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: In that test case, we cover the case if the input `getDoughs()` is `{}`, right?

src/tests/unit/doughs-dao.test.js
Ruefa: I would argue that unit tests should be used to test individual chunks of code in a vacuum. I think testing an "interface" of a class is better done by the integration tests. I think it's a trap to think that a method like `doughColumnAliases` doesn't need to be tested because you should be writing tests that catch errors now and _in the future_. I would agree though that `doughColumnNames` does not need an individual test because it literally returns constant values and doesn't really have logic like `doughColumnAliases` does. Another thing to keep in mind is that our code uses external dependencies that change on a regular basis. Lodash could decide to change how `_.toPairs()` works and we would want our tests to catch that.

jemisonf: But, if we can show that our function always calls `connection.execute()` with a valid query, does it matter whether or not our sql generator generates the correct query under every circumstance? Because the sql generator is unexported, any case where a user can pass it parameters to generate invalid SQL will be a case we can replicate by testing the interface of the exported functions.

tsoliangwu0130: Currently, the process of `getDoughs` is:

1. Take raw filters and normalize it.
2. Use the normalized filter to generate SQL query.
3. Call `connection.execute()` and execute with the generated SQL query.

Since the logic is pretty straightforward so far, only testing the step 3 should also cover the logic of step 1-2 to a certain degree. However, what if the process changed a little bit to something like this:

1. Take raw filters and normalize it.
2. Use the normalized filter to generate SQL query.
3. Call another helper function (called function A) to optimize the query (e.g. check if we can remove some unnecessary conditions or use `JOIN` clause to replace some subselect queries).
4. Call `connection.execute()` and execute with the optimized SQL query.

Assume there is a bug in step 1-2 and another one in step 3 (function A), but for some reasons these bugs coincidentally fix each other, what should we do? If you only test the final step, yes, the input query is what we expected, but does the testing process really cover what we should test?  Let's say if you add another function has a very similar logic in the future:

1. Take raw filters and normalize it.
2. Use the normalized filter to generate SQL query.
3. Call another helper function (called function B) to optimize the query.
4. Call `connection.execute()` and execute with the optimized SQL query.

If the function B is bug-free, theoretically the unit test will fail at step 4 since there's a bug in step 1-2, but how can you tell the bug is actually in step 1-2 instead of in the step 3 (function B)? If there is more than one step between 2-4, it'll be even harder to tell where the issue is, right?

kalcicd: I am undecided on this issue myself. Generally, I think of a 'unit' as an exported module because the exported unit can be a dependency for any number of other functions, thus should be tested. With helper functions, they are only called through their parent function, thus act more as a way to simplify code rather than to act as its own 'unit'. That being said, I do see the value in writing unit tests for helper functions. As Liang and Alex have brought up, if there is a bug or change in a helper that breaks the module, an additional unit test would allow the issue to be identified and fixed much more easily. Ultimately, I think that it is worth it to write unit tests for helper functions, as long as they are non trivial.

ianbrown9475: Personally, I don't think it makes sense to write unit tests for `doughColumnNames` or `doughColumnAliases`. These functions are essentially private methods to the dough dao class and I don't think it makes sense to test private methods in general.

When writing tests for methods in a module, there's a reason you aren't able to stub/mock other methods in that same module: you would essentially be modifying the module under test. If we were to undo the refactoring of `doughColumnAliases` and directly paste the logic in place of each call to it, there would be no functional difference in how the code works, since this function isn't being exported and used by other modules. But in this case, we obviously wouldn't want to modify those lines of code for testing, even if we could do that. Similarly, we wouldn't want to refactor arbitrary lines of code like [this](https://github.com/osu-mist/pizza-api/blob/b0b9e0753417a8751e4d1c9ace6ece02610721da/src/api/v1/db/oracledb/doughs-dao.js#L88-L92) into methods so that we could export them for testing, even if these tests might provide some value.

I think the reason for this is that if we later modify the dao module, even if the results of the exported functions are still correct, the unit tests could fail due to renaming, removing, or rewriting helper functions. We would also be obligated to add more tests for new helper functions. To me, this seems like unnecessary maintenance. I could also see this discouraging refactoring.

@tsoliangwu0130 from your previous example, I see how having unit tests on the helper functions could provide value by helping to pinpoint errors, but I don't think that's the only metric we should go off of. I think this would start overloading the purpose of unit tests from testing public method correctness -> testing public method correctness + debugging. In your example, if we are in the first process, even if the helper function is incorrect, the exported function is correct, which is what we're testing. As soon as we transition to the second process, we have an exported function that's incorrect, which would mean we have a failing unit test. I think it would be up to us at that point to debug the failing exported function.

If these kinds of helper functions start becoming large or complex, we probably want to move them to a separate module to avoid bloating the original module. At that point, we would want to add unit tests for that new module, since it's distinct from the original. But for simple helper functions, I don't think it makes sense to unit test them. Not as important, but it would also require us to add it to our exports and probably have to comment somewhere that the method is only being exported for testing to avoid confusion.

jemisonf: @ianbrown9475 really well put. I would agree that the case @tsoliangwu0130 suggested would require a refactor of the test code, since it significantly increases the complexity of the module internals. However, I think that refactor could be as simple as stubbing the SQL optimizer and testing which inputs we pass to that stub while also testing that we pass the stub's return on to the execute function.

src/tests/unit/doughs-dao.test.js
jemisonf: That's true, I am going to add a test to make sure it returns the output of SerializeDoughs even with different inputs.

src/tests/unit/doughs-dao.test.js
jemisonf: I am wondering if code coverage metrics would help resolve this at all. It seems like one good indicator of an unexported method that needs to be tested on its own might be if it's difficult to cover most of the lines of the method by just testing the module's exported methods. Not very familiar with common javascript coverage tools though.

jemisonf: Here's the output of `nyc --reporter=lcov --reporter=text gulp test`, per Ian's suggestion:
```
------------------------------------------------------|----------|----------|----------|----------|-------------------|
File                                                  |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------------------------------------|----------|----------|----------|----------|-------------------|
All files                                             |     76.7 |    42.11 |    54.55 |    77.58 |                   |
 pizza-api                                            |    88.46 |      100 |    57.14 |      100 |                   |
  gulpfile.js                                         |    88.46 |      100 |    57.14 |      100 |                   |
 pizza-api/dist/api/v1/db/oracledb/api/v1/db/oracledb |    80.28 |       50 |    68.75 |    80.88 |                   |
  connection.js                                       |       50 |    16.67 |        0 |       52 |... 52,54,55,57,58 |
  doughs-dao.js                                       |    97.78 |      100 |    91.67 |    97.67 |               130 |
 pizza-api/dist/api/v1/serializers/api/v1/serializers |    78.57 |      100 |        0 |    78.57 |                   |
  doughs-serializer.js                                |    78.57 |      100 |        0 |    78.57 |          22,23,31 |
 pizza-api/dist/tests/unit/tests/unit                 |      100 |      100 |      100 |      100 |                   |
  test-data.js                                        |      100 |      100 |      100 |      100 |                   |
 pizza-api/dist/utils/utils                           |    66.67 |    39.29 |    33.33 |    66.13 |                   |
  jsonapi.js                                          |       25 |        0 |        0 |       25 |... 52,60,62,73,81 |
  load-openapi.js                                     |       80 |      100 |      100 |       80 |             22,23 |
  logger.js                                           |    64.29 |        0 |        0 |    64.29 |    49,55,66,68,70 |
  paginator.js                                        |      100 |    91.67 |      100 |      100 |                13 |
  uri-builder.js                                      |    84.62 |        0 |       50 |    83.33 |             32,33 |
------------------------------------------------------|----------|----------|----------|----------|-------------------|
```

src/tests/unit/doughs-dao.test.js
jemisonf: I added proxyquire in the latest pr 👍

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: @jemisonf using code coverage metrics sounds probably a good idea, but I'm not sure which one should we use. The one you shared looks like a popular one, but I think we still need to do some more investigation.

@ianbrown9475 hmm, shouldn't [the purpose of the unit test](http://softwaretestingfundamentals.com/unit-testing/) includes debugging? I guess to me it's not important if the current unit is whether private or exported, the main point of writing unit tests is to help us easily detect and locate the issues/bugs if there's any, and even further to allow us easily extend our current logic.

But anyway, overall that sounds fair to me and I also agreed we shouldn't unit test those are trivial. I'd suggest let's discuss this on our team meeting tomorrow and see what other's opinions are.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,100 @@
...
+        await doughsDao.getDoughs(getDoughsData.multipleFilters);
+        connectionSpy
+          .getCall(0)
+          .calledWith(
+            getDoughsData.getDoughsQueryWithMultipleConditions,
+            getDoughsData.multipleFiltersBindParams,
+          ).should.be.true;
+      });
+    });
+  });
```
tsoliangwu0130: Same comments for these two test cases, since the input of `doughsDao.getDoughs` are different, although the result could be the same, I guess it's still better to asset the result explictly.

src/tests/unit/doughs-dao.test.js
jemisonf: Ah yeah, that's from `doughs-dao` generating the base url internally instead of having the serializer generate the base url, which seems to be the more common practice. I'll fix that.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,100 @@
...
+  let getConnectionStub;
+  let connectionSpy;
+  let doughsDao;
+  beforeEach(() => {
+    connectionSpy = sinon.stub().returns(getDoughsData.executeReturn);
+    serializeDoughsStub = sinon.stub().returns(getDoughsData.baseGetDoughsReturn);
+    getConnectionStub = sinon.stub().resolves({
+      execute: connectionSpy,
+      close: () => { },
+      commit: () => { },
```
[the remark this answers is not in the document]
jemisonf: We probably don't, I copied that idea from somewhere else 😬

src/tests/unit/test-data.js
```diff
@@ -0,0 +1,31 @@
+const getDoughsConditions = 'ID AS "id", NAME AS "name", GRAMS_FLOUR AS "gramsFlour", GRAMS_WATER AS "gramsWater", FLOUR_TYPE AS "flourType", WATER_TEMP AS "waterTemp", GRAMS_YEAST AS "gramsYeast", GRAMS_SALT AS "gramsSalt", GRAMS_SUGAR AS "gramsSugar", GRAMS_OLIVE_OIL AS "gramsOliveOil", BULK_FERMENT_TIME AS "bulkFermentTime", PROOF_TIME AS "proofTime", SPECIAL_INSTRUCTIONS AS "specialInstructions"';
+
```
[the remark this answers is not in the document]
jemisonf: I should be able to break the string into multiple lines without changing the query generator, I'll look into that.

src/tests/unit/doughs-dao.test.js
```diff
@@ -0,0 +1,100 @@
...
+    describe('when it has an invalid filter', () => {
+      it('does not parse the invalid filter', async () => {
+        await doughsDao.getDoughs(getDoughsData.invalidFilters);
+        connectionSpy
+          .getCall(0)
+          .should.have.been
+          .calledWith(
+            getDoughsData.getDoughsQuery,
+            getDoughsData.emptyBindParams,
+          );
```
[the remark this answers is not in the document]
jemisonf: I could see the argument for that. My thought originally was that if our tests show that it returns the output of serializeDoughs under two different cases it wouldn't be necessary to test all of them, but I'm not really committed to that. Maybe I should refactor the tests for this to iterate over a bunch of different input options and test each one to see if it still returns the right input.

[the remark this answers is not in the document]
jemisonf: The updated lines 55-66 have some logic to loop over all the different input filter categories and see if it still returns the output of serializeDoughs. Let me know what you think about that solution.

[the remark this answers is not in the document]
tsoliangwu0130: Looks good. The only suggestion is you can destructure the test data at once instead of accessing via getDoughsData every time since they're used many times during this unit test.

package.json
```diff
@@ -16,6 +16,7 @@
   "scripts": {
     "babel": "./node_modules/.bin/gulp babel",
     "build": "./node_modules/.bin/gulp build",
+    "coverage": "nyc --reporter=lcov --reporter=text ./node_modules/.bin/gulp test",
```
jemisonf: adding a `coverage` command because it came in handy during review earlier

src/tests/unit/doughs-dao.test.js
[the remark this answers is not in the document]
jemisonf: Thanks, good suggestion

jemisonf:
@tsoliangwu0130 thanks, that was a good catch. It looks like stubbing the `validateOracleDb` method from `connection` fixed the problem. Not entirely sure why--similar code in the terms api has the warning about missing configuration, but it doesn't cause a test to fail.

jemisonf:
Ok, I have now *actually* fixed this by stealing a line from `terms-dao` and using Sinon to stub the `config.get` method.

PR #7, opened 2019-11-05
CO-1626: adds POST /doughs endpoint

jemisonf:
Adds postDough method to doughs-dao
Adds serializeDough method to doughs-serializer
Updates method signature used in doughs path code


src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -49,6 +106,14 @@ const doughColumnAliases = _.toPairs(doughColumnNames)
 const doughsGetQuery = (conditionals) => `SELECT ${doughColumnAliases} FROM DOUGHS ${conditionals ? `WHERE ${conditionals}` : ''}`;
 
 
+/**
+ * A query to create a new dough
+ * @constant
+ * @type {string}
+ */
+const doughsPostQuery = `INSERT INTO DOUGHS (${doughColumns}) VALUES (${doughValues}) `
+  + `RETURNING ${doughsOutBindParamColumnNames} INTO ${doughsOutBindParamValues}`;
```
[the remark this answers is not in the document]
jemisonf: My take has historically been that this:
```js
const doughsPostQuery = `INSERT INTO DOUGHS (${doughColumns}) VALUES (${doughValues}) `
  + `RETURNING ${doughsOutBindParamColumnNames} INTO ${doughsOutBindParamValues}`;
```
is preferable to this:
```js
const doughsPostQuery = `INSERT INTO DOUGHS (${doughColumns}) VALUES (${doughValues}) \
RETURNING ${doughsOutBindParamColumnNames} INTO ${doughsOutBindParamValues}`;
```

For purely aesthetic reasons (and on frontend, I think they get compiled into the same thing). However, I checked the Airbnb style guide and their recommendation is:

> Strings that cause the line to go over 100 characters should not be written across multiple lines using string concatenation.

From [here](https://github.com/airbnb/javascript#strings--line-length). Meaning they would recommend not breaking the string at all.

Thoughts?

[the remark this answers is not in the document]
ianbrown9475: We don't follow the style guide's recommendation on this. Strings like this: https://github.sig.oregonstate.edu/ecs-data/students-api-contrib/blob/1679b47cb3e4b865b231ec6de8342ac0d8f30733/contrib.js#L275-L360 just wouldn't be feasible to keep on one line. We usually follow the same conventions as this example for formatting and readability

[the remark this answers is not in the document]
jemisonf: Ah fair enough, I can definitely see how regular multiline strings are superior for a query of that size. I'll update mine to match.

jemisonf:
@ianbrown9475 as a followup to your comment here:
> I'm pretty sure express-openapi handles default parameter values for us, so this shouldn't be necessary. Also, it's our convention to reject a request when the json body has unexpected properties. Instead of silently discarding them here, I would modify the openapi document to disallow them.

I refactored the validation logic so that the doughs `post` functions calls a `validateBody` util function to check the validity of the fields in the request, so theoretically only valid data is ever passed into `postDough`. This isn't 100% final but I'm curious to hear your thoughts on the approach overall.

.eslintrc.yml
```diff
@@ -47,6 +56,7 @@ rules:
     - publicOnly: true
       require:
         ArrowFunctionExpression: true
+  max-classes-per-file: 'off'
```
jemisonf: had to disable this to be able to define the error classes in `utils/validate-parameters`. there's possibly a more elegant solution but this doesn't seem to be a very useful rule for us anyways since mostly we're not using classes.

ianbrown9475:
@jemisonf The logic you implemented in `utils/validate-parameters.js` looks good, but this validation is already being done by `express-openapi`. When we call `initialize()` in `src/app.js`, some middleware functions are being set up by that package. These functions will appear earlier in the middleware chain than the path handlers. https://github.com/kogosoftwarellc/open-api/tree/master/packages/express-openapi

jemisonf:
@ianbrown9475 it looks like checking for the presence of the `data` key is already done, but the rest is not.

ianbrown9475:
@jemisonf So the validation for the `data` property is being done because you specify that it's required in your openapi file. `attributes` isn't marked as required, so it's allowed to be missing. Also, `additionalProperties` defaults to `true` in the openapi schema so since it isn't set to `false`, additional properties are allowed on your post body. If you want to change this behavior, you should just be able to change the openapi file and the rest will be handled by the package.

jemisonf:
@ianbrown9475 that makes sense. Getting that fixed now.

jemisonf:
> When I set name or flourType to "", I get a 500. When I set specialInstructions to "", the request goes through but it converts it to null

For name and flourType, I think the move here is to use a `minLength` of 1 for the strings. For `specialInstructions`, we have the limitation that oracledb can't actually store an empty string, and will simply always use null. I'm not sure what the best strategy to deal with that is.

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -117,37 +188,59 @@ const getDoughs = async (filters) => {
...
+ * It's assumed that body has already been checked for invalid properties meaning that when
+ * one is found here it's a server error rather than a user error.
+ */
+const getPostBindParams = (body) => {
+  const { attributes } = body.data;
+  const bindParams = doughsOutBindParams;
+  _.forEach(attributes, (attributeValue, attributeName) => {
+    if (!(attributeName in doughsProperties)) {
+      throw new Error('Invalid attribute found');
+    }
```
jemisonf: I don't know if this code should trust 100% that the openapi doc will catch every possible invalid param; this seems like some good defense in depth to stop any errors before they hit the database

ianbrown9475: That's probably fine for this case since the bind parameters are being generated programmatically. Usually these end up being hardcoded instead. Could you also update the documentation for this function?

jemisonf:
> I think the easiest way to deal with "" for specialInstructions is to just store it as NULL in the database and convert it to an empty string in responses

This is a good idea. I'm using a `jsonapi-serializer` `transform` function for this, which I added as an option to our `serializerOptions` helper function.

src/api/v1/db/oracledb/doughs-dao.js
jemisonf: @ianbrown9475 what did you mean by updating the documentation? It already has an `@throws` note and I don't think the signature has changed. Sorry if I'm missing something.

src/utils/jsonapi.js
```diff
@@ -47,6 +48,7 @@ const serializerOptions = (serializerArgs) => {
       },
     },
     topLevelLinks: { self: topLevelSelfLink },
+    transform: transformFunction,
```
ianbrown9475: Nice, this seems like it would be helpful. Could you add a default value for this property though?

jemisonf: Spent way too much time thinking about if `false` or `null` was the better option for the default. Going with `null` because it's sufficiently false-y to stop [this check](https://github.com/SeyZ/jsonapi-serializer/blob/master/lib/serializer-utils.js#L274) and it matches better with the expected type of `transformFunction`.

ianbrown9475: I agree `null` would be better than `false`, although I could see this causing issues if the maintainer decided to check if `opts.transform === undefined`, which is fairly common for checking if a property is on an object or not while allowing other falsey values. I think [`_.identity`](https://lodash.com/docs/4.17.15#identity) might be better here

jemisonf: Hmm, if `undefined` is the more common false-y value to use here then it might make sense to leave it without a default, since that's what it ends up being if we don't give it an alternative value.

jemisonf: or just use `transformFunction || undefined`, although that feels too verbose.

ianbrown9475: I'm not sure if this is ideal either, since there are other ways to check if an object has a property like `hasOwnProperty` that would return different results if an object didn't have a property than if it had the property with value `undefined`. We should be able to make this work regardless of implementation of the jsonapi serializer package. The two main ways I can think of is to omit the property from the object that we pass in or use the identity function.

jemisonf: What would the usage of `identity` look like in this case? I was looking at the docs but I'm not totally clear about what you have in mind.

ianbrown9475: From what I understand, `_.identity` is just a name for the function `(x) => x`. So if we use the identity function as the transformation function, we're essentially saying that we aren't changing anything.

jemisonf: Ah, I see what you mean. My concern there is adding the additional overhead of making an unnecessary function call for potentially every record in a large array of results, even if it is a really simple function. Although I wouldn't be surprised if the interpreter were to optimize away the function call in the case that `identity` was being used.

I think that finding a way to just not include the attribute when no `transformFunction` is passed would be a better option.

jemisonf: I will look into this when I get in tomorrow 👍

ianbrown9475: Good point. I would agree that omitting the optional property is better than explicitly setting it with something designed to not do anything.

jemisonf:
Given a bunch of options for doing conditional object property assignment, I went with what I thought was the clearest:
```js
const options = {
    pluralizeType: false,
    attributes: resourceKeys,
    id: identifierField,
    keyForAttribute: keyForAttribute || 'camelCase',
    dataLinks: {
      self: (row) => {
        if (enableDataLinks) {
          return resourcePathLink(resourceUrl, row[identifierField]);
        }
        return null;
      },
    },
    topLevelLinks: { self: topLevelSelfLink },
  };

  if (transformFunction) options.transform = transformFunction;
```

This was another candidate, but imo it's less readable:
```js
const options = {
    ...{
      pluralizeType: false,
      attributes: resourceKeys,
      id: identifierField,
      keyForAttribute: keyForAttribute || 'camelCase',
      dataLinks: {
        self: (row) => {
          if (enableDataLinks) {
            return resourcePathLink(resourceUrl, row[identifierField]);
          }
          return null;
        },
      },
      topLevelLinks: { self: topLevelSelfLink },
    }
    ...(transformFunction) && {transform: transformFunction},
  };
```

PR #8, opened 2019-11-07
CO-1633: implement GET /ingredients

jemisonf:


jemisonf:
Because the logic for processing the filters was _so_ similar to `GET doughs`, I pulled it out into a util module that I'm currently using just for ingredients. I did not do the same for doughs because that would require rewriting the tests there and it seemed out of scope of this ticket, but that would be a definite next step.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'lodash';
+import { openapi } from 'utils/load-openapi';
+import { serializeIngredients } from 'api/v1/serializers/ingredients-serializer';
+import { GetFilterProcessor } from 'utils/process-get-filters';
+import { getConnection } from './connection';
```
tsoliangwu0130: Take a look at our style guide here: https://github.com/osu-mist/style-guides/blob/master/JavaScript.md#import-order

> Whenever possible, the following should be grouped together, separated by a blank line at the top of a module in this order

> Each line within groups 1-3 should be ordered alphabetically by the first alphanumeric character in the name of the package/file in the statement.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -0,0 +1,56 @@
...
+  id: 'ID',
+  ingredientType: 'TYPE',
+  name: 'NAME',
+  notes: 'NOTES',
+};
+
+/**
+ * SQL aliases for mapping INGREDIENTS column names to
+ * IngredientResource property names
+ * @const {string}
```
tsoliangwu0130: You probably want a blank line between description and `@const` as you did in dough: https://github.com/osu-mist/pizza-api/blob/master/src/api/v1/db/oracledb/doughs-dao.js#L11

Same comment with your other JSDoc sections.

src/utils/process-get-filters.js
```diff
@@ -0,0 +1,60 @@
+import _ from 'lodash';
+/**
```
tsoliangwu0130: Need a blank line between import section and the JSDoc section.

src/utils/process-get-filters.js
```diff
@@ -0,0 +1,60 @@
...
+ * Removes the "filter[]" wrapper from a filter parameter
+ *
+ * @param {string} filterName
+ * @returns {string} the filter name with 'filter' and brackets removed
+ */
+const normalizeFilterName = (filterName) => {
+  const filterTextRegex = /filter\[(.*)\]/g;
+  const regexResults = filterTextRegex.exec(filterName);
+  return regexResults ? regexResults[1] : filterName;
```
tsoliangwu0130: Or even shorter:

```js
return regexResults[1] || filterName;
```

src/utils/process-get-filters.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/**
+ * Returns filters with keys changed to remove the "filter[]" wrapper
+ *
+ * @param {object} filters
+ * @returns {object}
+ */
+const normalizeFilterNames = (filters) => _.mapKeys(filters,
```
tsoliangwu0130: Not really important, but I guess I'll make `normalizeFilterNames` and `normalizeFilterName` be static methods of `GetFilterProcessor` class.

src/utils/process-get-filters.js
jemisonf: ooh, smart!

src/utils/process-get-filters.js
jemisonf: my goal was to make them private but I think you're right that having them be static is overall the better approach

src/utils/process-get-filters.js
```diff
@@ -1,36 +1,42 @@
...
+    const regexResults = filterTextRegex.exec(filterName);
+    return regexResults[1] || filterName;
+  }
+
   /**
    * Initializes a new filter process using the parameters from the
    * get request for the resource and a map of column names to resource attribute names
-   * @param {array} getParameters
+   *
+   * @param {Array} getParameters
```
tsoliangwu0130: Array of?

jemisonf: changed => `Array<object>`, would be handy to have typescript types here :/

tsoliangwu0130: And it should be `object[]`: https://github.com/osu-mist/students-api/blob/207194b7d9fa346494972ef2fc14773f8e72ee7b/src/utils/jsonapi.js#L20

jemisonf: this should be actually fixed and the commits should be pushed now

PR #9, opened 2019-11-18
CO-1641: add GET /ingredients unit tests

jemisonf:
Adds unit tests for `process-get-filters` and `ingredients-dao`. Merge after #8


src/tests/unit/ingredients-dao.test.js
```diff
@@ -0,0 +1,145 @@
...
+      ingredientsDao = proxyquire('../../api/v1/db/oracledb/ingredients-dao', {
+        './connection': {
+          getConnection: getConnectionStub,
+        },
+        '../../serializers/ingredients-serializer': {
+          serializeIngredients: serializeIngredientsStub,
+        },
+        '../../../../utils/process-get-filters': {
+          GetFilterProcessor: getFilterProcessorContructorStub,
+        },
```
jemisonf: This is stubbed with proxyquire in only this test because stubbing objects in sinon is, as it turns out, a pain. Because the constructor has no side effects outside of the class and the only public method of the class is stubbed, there's no functional difference between this approach and one that more thoroughly stubs out the class in every function, besides that this approach is more concise.

src/tests/unit/process-get-filters.test.js
```diff
@@ -0,0 +1,133 @@
...
+    it('returns an empty object', () => {
+      const normalizedFilters = GetFilterProcessor.normalizeFilterNames({});
+      normalizedFilters.should.deep.equal({});
+    });
+  });
+  describe('when it gets a non empty object', () => {
+    it('changes filter keys but leaves non-filter keys unchanged', () => {
+      const normalizedFilters = GetFilterProcessor.normalizeFilterNames({
+        ...fooFilter,
+        ...nameFilter,
```
jemisonf: I liked this approach for composing multiple test data objects together because I think it's pretty clear what it's doing and it lets me write less test data code overall. Open to suggestions though.

src/tests/unit/ingredients-dao.test.js
```diff
@@ -0,0 +1,145 @@
...
+    testConnectionReturnRows,
+  } = getIngredientsData;
+  let serializeIngredientsStub;
+  let getConnectionStub;
+  let connectionSpy;
+  let ingredientsDao;
+  let getFilterProcessorStub;
+  let getFilterProcessorContructorStub;
+  beforeEach(() => {
+    connectionSpy = sinon.stub().returns(testConnectionReturn);
```
ianbrown9475: Not a big deal, but could you name this something like `connectionStub` instead since stubs are somewhat different from spies?

src/tests/unit/ingredients-dao.test.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+    getFilterProcessorContructorStub = sinon.stub();
+    getFilterProcessorStub = sinon.stub(GetFilterProcessor.prototype, 'processGetFilters').returns({
+      bindParams: testBindParams,
+      conditionals: testConditionals,
+    });
+
+    Object.setPrototypeOf(GetFilterProcessor, getFilterProcessorContructorStub);
+
+    ingredientsDao = proxyquire('../../api/v1/db/oracledb/ingredients-dao', {
```
ianbrown9475: For this first path passed to `proxyquire`, you can just use `'api/v1/...'` since you want the path to evaluate relative to the current module. It's just the path keys in the object you pass that need to be relative to the proxyquired module.

src/tests/unit/ingredients-dao.test.js
```diff
@@ -0,0 +1,145 @@
...
+  });
+  describe('when the constructor for GetFilterProcessor is called', () => {
+    beforeEach(() => {
+      ingredientsDao = proxyquire('../../api/v1/db/oracledb/ingredients-dao', {
+        './connection': {
+          getConnection: getConnectionStub,
+        },
+        '../../serializers/ingredients-serializer': {
+          serializeIngredients: serializeIngredientsStub,
+        },
```
ianbrown9475: Aren't these stubs already being taken care of in the top-level `beforeEach()`?

src/tests/unit/ingredients-dao.test.js
```diff
@@ -0,0 +1,145 @@
...
+        '../../../../utils/process-get-filters': {
+          GetFilterProcessor: getFilterProcessorContructorStub,
+        },
+      });
+    });
+    it('passes the right values to the GetFilterProcessor constructor', async () => {
```
ianbrown9475: Does this function need to be async?

src/tests/unit/process-get-filters.test.js
```diff
@@ -0,0 +1,133 @@
...
+    nameBindParams,
+    nameConditional,
+    bahFilter,
+    emptyBindParams,
+    emptyConditional,
+  } = processGetFiltersData;
+  beforeEach(() => {
+    processor = new GetFilterProcessor(
+      [
+        fooParamName, nameParamName,
```
ianbrown9475: Could you put a newline between these two since the array is split over multiple lines?

src/tests/unit/process-get-filters.test.js
```diff
@@ -0,0 +1,133 @@
...
+    it('generates bind params and conditionals only from the valid filter', () => {
+      const { bindParams, conditionals } = processor.processGetFilters({
+        ...fooFilter,
+        ...bahFilter,
+      });
+      bindParams.should.deep.equal(fooBindParams);
+      conditionals.should.equal(fooConditional);
+    });
+  });
+  describe('when it gets filters that do not user the `filter[...]` syntax', () => {
```
ianbrown9475: Typo user -> use

src/tests/unit/test-data.js
```diff
@@ -37,6 +37,77 @@ const getDoughsData = {
...
+  nameConditional: 'NAME = :name',
+  fooBahFilters: {
+    bah: 'humbug',
+    'filter[foo]': 'baz',
+  },
+  fooNameFilters: {
+    name: 'shrek',
+    'filter[foo]': 'baz',
+  },
+
```
ianbrown9475: Could you remove this blank line?

src/tests/unit/doughs-dao.test.js
```diff
@@ -7,6 +7,8 @@ import sinonChai from 'sinon-chai';
 
 import { getDoughsData } from './test-data';
 
+sinon.restore();
```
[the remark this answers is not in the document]
jemisonf: Yeah, I was noticing that `sinon.replace(config, 'get', () => ({ oracledb: {} }));` was causing an error because it was used in multiple modules

src/tests/unit/ingredients-dao.test.js
jemisonf: left a comment about that [here](https://github.com/osu-mist/pizza-api/pull/9/files/873777b02cfaa8aaaa65bdbf84cd67d60d513171#r347666661). This `proxyquire` call is stubbing out `GetFilterProcessor` in addition to the normal stubs because we need to examine the interaction with the constructor, but it's left out in other tests because the constructor doesn't have side effects and it's much simpler to not stub it in those cases.

src/tests/unit/process-get-filters.test.js
jemisonf: might move these to just all be on one line instead since it's a short array

src/tests/unit/process-get-filters.test.js
```diff
@@ -0,0 +1,133 @@
...
+
+chai.should();
+
+describe('normalizeFilterNames', () => {
+  const {
+    fooFilter, normalizedFooFilter,
+    nameFilter, normalizedNameFilter,
+  } = processGetFiltersData;
+  describe('when it gets an empty object', () => {
+    it('returns an empty object', () => {
```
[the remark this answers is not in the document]
jemisonf: yeah, it creates a nice visual hierarchy to give you context when a test fails too

jemisonf:
Added some additional comments in the test code. Let me know if there's anything I missed that you think needs documentation.

src/tests/unit/doughs-dao.test.js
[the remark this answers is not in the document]
ianbrown9475: I think what's happening here is that when the tests are about to be run, they are all imported at the same time. Meaning these top-level replaces/restores are all occurring at once. This isn't ideal since tests are affecting each other. Could you fix this?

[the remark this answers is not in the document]
jemisonf: Looks like moving `sinon.replace` to the top level `beforeEach` call fixes it, unsurprisingly. I do believe, for what it's worth, that Mocha executes tests in serial which is why this didn't cause any errors as is.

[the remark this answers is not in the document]
ianbrown9475: Right, the tests are executed serially, but all of the modules are required/imported at once. So when one of the modules had already called replace, if another module called replace without first calling restore, there would be an error since the the replacement had already occurred. Also, the config replacement was only lasting until the first call to restore (in `afterEach`) which isn't good since we always want this to be stubbed out. Moving them to a beforeEach is the best solution, since the stub will exist for all of the tests without causing any race conditions.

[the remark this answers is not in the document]
jemisonf: Fair. I can see a lot of ways in which this is the best solution.

src/tests/unit/doughs-dao.test.js
```diff
@@ -61,17 +59,19 @@ describe('test doughs dao', () => {
...
+  afterEach(() => {
+    sinon.restore();
+  });
+  describe('getDoughs', () => {
+    beforeEach(() => {
+      connectionStub = sinon.stub().returns(executeReturn);
 
-        doughsDao = proxyquireDoughsDao(connectionStub, serializeDoughsStub, serializeDoughStub);
-      });
+      doughsDao = proxyquireDoughsDao(connectionStub, serializeDoughsStub, serializeDoughStub);
```
jemisonf: The changes in this section made for a really messy rebase. I've verified that it works though and the final changes look OK to me.

PR #10, opened 2019-11-20
CO-1642: implement POST /doughs unit tests

jemisonf:


src/tests/unit/doughs-dao.test.js
```diff
@@ -110,12 +129,54 @@ describe('test doughs dao', () => {
...
+    });
+  });
+
+  describe('postDoughs', () => {
+    beforeEach(() => {
+      connectionStub = sinon.stub().returns(testDbReturn);
+
+      doughsDao = proxyquireDoughsDao(connectionStub, serializeDoughsStub, serializeDoughStub);
+    });
+    describe('when it gets valid params', () => {
```
jemisonf: There's not a lot of logic in these tests, reflecting the fact that there's not a lot of logic in the code itself. I'm open to suggestions for alternative approaches, but I wanted to be careful about spending time testing cases where behavior is "undefined" (like openapi not correctly filling in defaults, or not generating defaults when parameters are missing, or not validating requests properly) because my assumption when developing this was that we could basically totally trust openapi validation. 

The exception to that is the case where we get an invalid parameter, since I would be concerned about the potential for SQL injection or other security issues if that occurred, and because the behavior is basically "defined" (ie the code throws an exception) anyways.

Open to suggestions, but that's my basic reasoning here.

PR #11, opened 2019-11-20
CO-1648: implement POST /ingredients

jemisonf:


src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
   notes: 'NOTES',
 };
 
+/**
+ * convert an output bind param name like `nameOut` to a
+ * property name like `name` by removing 'Out' at the end of the string
+ * @param {string} outBindParamName
+ * @returns {string} the property name
+ */
+const outBindParamToPropertyName = (outBindParamName) => outBindParamName.slice(0, -3);
```
tsoliangwu0130: I'd suggest using regex to fetch the substring you need instead of using index to slice the string.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
   notes: 'NOTES',
 };
 
+/**
+ * convert an output bind param name like `nameOut` to a
+ * property name like `name` by removing 'Out' at the end of the string
+ * @param {string} outBindParamName
```
tsoliangwu0130: Can you add a blank line between description and `@param`.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
...
+ * prepended with `:` and joined with commas
+ * @constant
+ * @type {string}
+ */
+const ingredientValues = _.map(ingredientsProperties, (propertyValues, property) => `:${property}`)
+  .join(', ');
+
+/**
+ * The "out" bind params for a POST/PATCH request to INGREDIENTS.
+ *
```
tsoliangwu0130: This blank line should be moved below the next line.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
...
+const outBindParamToPropertyName = (outBindParamName) => outBindParamName.slice(0, -3);
+
+/**
+ * A list of the columns in the INGREDIENTS table joined by commas
+ * @constant
+ * @type {string}
+ */
+const ingredientColumns = _.map(ingredientsProperties,
+  (propertyValues, property) => ingredientsColumnNames[property])
+  .join(', ');
```
tsoliangwu0130: You can just use [`_.keys`](https://lodash.com/docs/4.17.15#keys) to improve the logic here.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
...
+
+/**
+ * The full list of valid ingredient attributes for a POST request,
+ * prepended with `:` and joined with commas
+ * @constant
+ * @type {string}
+ */
+const ingredientValues = _.map(ingredientsProperties, (propertyValues, property) => `:${property}`)
```
tsoliangwu0130: Same, you didn't use values of the object at all so I think you can improve it by using `_.keys` as well.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
...
+ * The column names referenced by the output bind params
+ * used when inserting a new ingredient, joined with commas
+ * @constant
+ * @type {string}
+ */
+const ingredientsOutBindParamColumnNames = _.map(ingredientsOutBindParams,
+  (bindParamValue, bindParamName) => ingredientsColumnNames[
+    outBindParamToPropertyName(bindParamName)
+  ])
+  .join(', ');
```
tsoliangwu0130: Same comment, try using _.keys to improve it.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -15,6 +17,70 @@ const ingredientsColumnNames = {
...
+
+/**
+ * The names of the out bind params used when inserting
+ * a new ingredient, joined with commas
+ * @constant
+ * @type {string}
+ */
+const ingredientsOutBindParamValues = _.map(ingredientsOutBindParams,
+  (paramValue, paramName) => `:${paramName}`)
+  .join(', ');
```
tsoliangwu0130: Same comment as above.

openapi.yaml
```diff
@@ -751,8 +751,10 @@ definitions:
         type: string
         description: additional information about using or preparing the ingredient
         example: brown the sausage in a pan before baking
+        default: ''
```
tsoliangwu0130: Just curious, why can't this be `null`?

src/api/v1/db/oracledb/ingredients-dao.js
jemisonf: If we used `_.keys` though, wouldn't we have to call `map` again on the result? I'm not sure that's a clear improvement over just `_.map`.

openapi.yaml
jemisonf: Because oracledb can't distinguish `null` from `''`, we don't have a way to store an empty string in a way that we can distinguish from a null value, so it's simpler to just only allow one of those values. That way, someone doesn't write code expecting an empty string and then getting `null`, or the other way around.

src/api/v1/db/oracledb/ingredients-dao.js
jemisonf: Hmm. I guess the advantage of a regex is that that function can more easily handle more edge cases, like getting a bind param name that doesn't end with `Out`. My thought when writing it was that because it's a private method we have pretty strict control over when it's used, so we can basically always assumed it's called with the "right" input. In that case, I thought this solution was simpler and more readable. Thoughts?

tsoliangwu0130: Just my two cents, I think using regex is easier for developers including maintainers to quickly understand which part of string should be fetched since it's self-descriptive. Counting the index of a string and slice the part we need sometimes can be "simpler", but I think it'll be harder to maintain as time goes by, and it'll be even harder if you have more similar logic like that (at least you'll need to write some input and output examples for them to let others know what should be parsed).

jemisonf: That makes sense. Thoughts on using `.replace(/Out$/, "")` instead of either solution? I feel like that strikes a balance of being readable and more error resilient.

src/api/v1/db/oracledb/ingredients-dao.js
tsoliangwu0130: You are right, I didn't notice you used the keys of as new keys `ingredientsProperties` of `ingredientsColumnNames`. Then yes, using `_.map` sounds like the best solution here. Good catch.

src/api/v1/db/oracledb/ingredients-dao.js
tsoliangwu0130: Another benefit of using regex is that you can also validate the input. I won't use replace in this case since it will not do anything if the input is incorrect. Instead, I'll use regex pattern to match the input and fetch the part we need by groups.

tsoliangwu0130: Like what you did [here](https://github.com/osu-mist/pizza-api/blob/0d1a11b94795c0bcb925d47421ba97dba9b96ece/src/api/v1/db/oracledb/doughs-dao.js#L130-L132).

openapi.yaml
tsoliangwu0130: Based on the discussion [here](https://stackoverflow.com/questions/13278773/null-vs-empty-string-in-oracle), I think Oracle will convert the empty strings to null by default, and that's expected. I guess in this case, we shouldn't say the default value of it is an empty string in swagger since it's misleading. You can just simply remove the defualt value and insert null if no value is provided.

tsoliangwu0130:
Looks like there's some conflict here after you merge other branch to master. Can you resolve it as well?

openapi.yaml
jemisonf: I don't know if that would be expected from the perspective of an API user. I'm also using a transform function in the serializer so that the API always returns an empty string, even though the database stores it as `null`. I think that's preferable to the case where someone passes in an empty string when they're creating an ingredient record and then gets `null` returned to them in the API even though they were expecting an empty string.

Ian and I discussed this [here](https://github.com/osu-mist/pizza-api/pull/7#pullrequestreview-316390831) with the PR for POST doughs if you want to see his take on it.

src/api/v1/db/oracledb/ingredients-dao.js
jemisonf: Got this updated to use a regular expression 👍

openapi.yaml
tsoliangwu0130: I'd argue with it. To me, the meaning of empty strings and null are slightly different. Using instruction as an example, an empty string can stand for there is no instruction at all, however, null means either there's no instruction or just not having yet.

tsoliangwu0130: @Ruefa I guess we need another tiebreaker lol, can you share your thought on it?

tsoliangwu0130: Plus, I think returning null is very common behavior, so I don't think it's necessary to convert the value to an empty string. We could just return whatever we saw in the database.

jemisonf: I guess the last thing I would add is that the openapi spec gives us a way to communicate that an attribute will always store an empty string if there's no value, but no way to communicate that an attribute will either be `null` or a string with length > 1. At least as far as I know.

jemisonf: Actually, the last last thing I will add is that in javascript an empty string mostly behaves like null, but null does not always behave like an empty string. So if you're writing frontend code at least, it's worse to get `null` when you're expecting an empty string than to get an empty string when you're expecting `null`.

src/api/v1/serializers/ingredients-serializer.js
```diff
@@ -34,4 +45,27 @@ const serializeIngredients = (rawIngredients, query) => {
   ).serialize(rawIngredients);
 };
 
-export { serializeIngredients };
+/**
+ * Serializes a single ingredient
+ * @param {object} rawIngredient
```
tsoliangwu0130: Blank line between description and @param.

jemisonf: I promise my next PR will have the right spacing in all the jsdoc comments 😢

PR #12, opened 2019-11-26
CO-1649: adds unit tests for POST /ingredients

jemisonf:
Like #11 , this is very similar to the unit tests for POST /doughs.


jemisonf:
Should be merged after #11

PR #13, opened 2019-11-26
CO-1656: unit test serializers

jemisonf:


src/tests/unit/doughs-serializer.test.js
```diff
@@ -0,0 +1,120 @@
...
+  emptyStringSpecialInstructionsDough,
+} = doughSerializerData;
+
+describe('test doughs serializer', () => {
+  before(() => {
+    sinon.replace(config, 'get', () => ({ oracledb: {} }));
+    // importing doughSerializer normally causes config to become unextensible,
+    // which then breaks other tests. This prevents that problem, even though
+    // stubbing imports is not required.
+    doughSerializer = proxyquire('../../api/v1/serializers/doughs-serializer', {});
```
jemisonf: This was a strange problem to have crop up. I can elaborate if necessary but this seems to have totally fixed it.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -190,7 +190,7 @@ const postIngredient = async (body) => {
       { autoCommit: true },
     );
     const rawIngredient = convertOutBindsToRawIngredient(result.outBinds);
-    return serializeIngredient(rawIngredient);
+    return serializeIngredient(rawIngredient, 'ingredients');
```
jemisonf: this + the changes in `ingredients-serializer` was a bug that my tests found!

src/tests/unit/doughs-serializer.test.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+  after(() => {
+    sinon.restore();
+  });
+  describe('serializeDough', () => {
+    describe('when it gets a dough object with specialInstructions = `null`', () => {
+      beforeEach(() => {
+        rawDough = nullSpecialInstructionsDough;
+      });
```
Ruefa: I see `beforeEach` used a lot and they are only slightly different each time. You can add a function to build the `beforeEach` to make your code a little less repetivive.
```
  const beforeEachMaker = (dough) => {
    beforeEach(() => {
      rawDough = dough;
    });
  };
```

src/tests/unit/doughs-serializer.test.js
```diff
@@ -0,0 +1,120 @@
...
+      it('adds all of the right attributes to `attributes`', () => {
+        const serializedDough = doughSerializer.serializeDough(rawDough, 'doughs');
+        _.forIn(doughAttributes, (value, attribute) => {
+          serializedDough.data.attributes.should.have.property(attribute);
+        });
+        serializedDough.data.attributes.should.not.have.property('id');
+      });
+    });
+  });
+  describe('serializeDoughs', () => {
```
Ruefa: Make use of whitespace to make your code more readable. A space between `serializeDough` and `serializeDoughs` tests will help segment the tests for others.

src/tests/unit/doughs-serializer.test.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+  after(() => {
+    sinon.restore();
+  });
+  describe('serializeDough', () => {
+    describe('when it gets a dough object with specialInstructions = `null`', () => {
+      beforeEach(() => {
+        rawDough = nullSpecialInstructionsDough;
+      });
+      it('converts specialInstructions to an empty string', () => {
```
Ruefa: This test case and the following one are almost the same. Make use of loops to reduce code reuse. Here is an example in my [learning project](https://github.com/osu-mist/game-curator-api/blob/4929a92be94da94b9fb2084382360c9832da033b/tests/unit/games-dao.test.js#L112-L133).

src/tests/unit/doughs-serializer.test.js
Ruefa: I ran into the same issue. If I remember right the problem is that importing `doughs-serializer` causes config to run. So you have to replace config before `proxyquire`.

src/tests/unit/ingredients-serializer.test.js
```diff
@@ -0,0 +1,128 @@
...
+  after(() => {
+    sinon.restore();
+  });
+  describe('serializeIngredient', () => {
+    describe('when it gets an ingredient object with `notes` = null', () => {
+      beforeEach(() => {
+        rawIngredient = nullNotesIngredient;
+      });
+      it('changes null to empty string', () => {
+        const serializedIngredient = ingredientSerializer
```
Ruefa: Some of the comments I left on the dough-serializer applies here as well. Especially looping tests. Look through your code and see where you can reduce code reuse. In fact it looks like some tests here are similar to the ones in dough-serializer. You can make a test-helper class to reuse test cases.

jemisonf:
Hey @Ruefa ,

I added some changes to the tests to reduce code reuse. My concerns with relying too much on shared libraries and test case loops is that it sacrifices some readability and you lose some of the clean idiomatic BDD style without significant performance gains, so I've mostly avoided those techniques. I'm open to reconsidering that though. Let me know your thoughts.

src/tests/unit/doughs-serializer.test.js
jemisonf: Yeah, that was my impression as well. Let me know if there's any issues with the approach I used here.

PR #14, opened 2019-12-02
CO-1647: add get dough by ID endpoint

jemisonf:


openapi.yaml
```diff
@@ -998,19 +998,19 @@ parameters:
...
     required: true
-    type: string
+    type: integer
     description: ID of pizza to retrieve
   doughId:
     in: path
     name: doughId
     required: true
-    type: string
+    type: integer
```
jemisonf: changing this and others to `integer` because our database column uses an integer and it simplifies the code a lot if we don't have to do the conversion ourselves

src/api/v1/paths/doughs/{doughId}.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Get a dough with id `req.params.doughId`
+ *
+ * @type {RequestHandler}
+ */
+const get = async (req, res) => {
+  try {
+    const result = await getDoughById(req.params.doughId);
+    if (result.data.length === 0) {
+      return errorBuilder(res, '404', `No dough with id ${req.params.doughId} exists`);
```
jemisonf: not sure if there's a standard text for this detail message, or if it's actually necessary

openapi.yaml
tsoliangwu0130: Based on the JSON API, ID *MUST* be a string: https://jsonapi.org/format/#document-resource-object-identification

src/api/v1/paths/doughs/{doughId}.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Get a dough with id `req.params.doughId`
+ *
+ * @type {RequestHandler}
+ */
+const get = async (req, res) => {
+  try {
+    const result = await getDoughById(req.params.doughId);
+    if (result.data.length === 0) {
```
tsoliangwu0130: No need to change, but I personally prefer using `!result.data` or [`_.isEmpty`](https://lodash.com/docs/4.17.15#isEmpty) to check if an array is empty.

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -245,4 +245,22 @@ const postDough = async (body) => {
...
+ * @param {number} id
+ * @returns {Promise<object>} the dough with id `id`
+ */
+const getDoughById = async (id) => {
+  const query = doughsGetQuery('ID = :id');
+  const bindParams = { id };
+  const connection = await getConnection();
+  try {
+    const { rows } = await connection.execute(query, bindParams);
+    return serializeDough(rows, `doughs/${id}`);
```
tsoliangwu0130: Here is an example I got for `doughs/{id}`:

```json
{
    "links": {
        "self": "https://localhost/v1/doughs/1"
    },
    "data": [
        {
            "type": "dough",
            "id": "1",
            "links": {
                "self": "https://localhost/v1/doughs/1"
            },
            "attributes": {
                "name": "Sample dough",
                "gramsFlour": "500",
                "flourType": "Bread flour",
                "gramsWater": "400",
                "waterTemp": "90",
                "gramsYeast": "20",
                "gramsSalt": "10",
                "gramsSugar": "0",
                "gramsOliveOil": "50",
                "bulkFermentTime": "120",
                "proofTime": "30",
                "specialInstructions": "Can also proof overnight in the refrigerator"
            }
        }
    ]
}
```

A couple issues here:

1. Integer fields are still using strings.
2. `data` is an array instead of a object.

You probably wanna check your `/doughs` again as well.

src/api/v1/paths/doughs/{doughId}.js
tsoliangwu0130: That looks fine to me, the only comment I have is you probably wanna change `id` to `ID` in text.

openapi.yaml
jemisonf: Hmm, I see why we had it that way originally then. It doesn't seem like other learning project APIs handle the case where a user passes in a string instead of an int, is that acceptable behavior here?

src/api/v1/db/oracledb/doughs-dao.js
jemisonf: Yep, I'm seeing that in `/doughs` as well. I'm not sure why that is, I'll investigate.

openapi.yaml
tsoliangwu0130: Which one? Looks like Parks API, Poker API and Game Curator API are all treating IDs as strings?

src/api/v1/paths/doughs/{doughId}.js
jemisonf: `!result.data` wasn't working for whatever reason, but `_.isEmpty` looks like a good idea.

tsoliangwu0130: Hmm, can you investigate why `!result.data` didn't work?

```js
> data = []
> !data
false
```

openapi.yaml
jemisonf: Sorry, meant to say it doesn't look like any of them handle the case where someone passes a string that can't be parsed to an integer, like `'abc'` or similar as opposed to `'1'` or `'2'`.

src/api/v1/paths/doughs/{doughId}.js
tsoliangwu0130: My bad, looks like [`[]` is trythy in JavaScripts](https://guide.freecodecamp.org/javascript/falsy-values/). Weird language.

openapi.yaml
tsoliangwu0130: Actually they did. They added [pattern](https://github.com/osu-mist/game-curator-api/blob/master/openapi.yaml#L94) in openapi file so that it will be validated automatically by the package we are using.

src/api/v1/paths/doughs/{doughId}.js
jemisonf: Yeah, that behavior actually kinda makes sense in my opinion but it's definitely not consistent with the rest of the language.

openapi.yaml
jemisonf: Oh, perfect! I'll go with that solution then.

src/api/v1/db/oracledb/doughs-dao.js
jemisonf: This fix fits right into the transform function. I'm surprised it's not something that the serializer can do automatically.

tsoliangwu0130: This is how I transformed strings to floating numbers in Students API in case you need it: https://github.com/osu-mist/students-api/blob/defffa5442f163cedcb1f63b0c87ff17923256e3/src/api/v1/serializers/students-serializer.js#L79-L85

jemisonf: Thanks, I like that idea of looping over a list of parameter names instead of writing each one out.

src/api/v1/serializers/doughs-serializer.js
```diff
@@ -19,6 +19,20 @@ const doughResourceUrl = resourcePathLink(apiBaseUrl, doughResourcePath);
...
+
+  const intFields = [
+    'gramsFlour',
+    'gramsWater',
+    'waterTemp',
+    'gramsYeast',
+    'gramsOliveOil',
+    'bulkFermentTime',
+    'proofTime',
+  ];
```
tsoliangwu0130: What about `gramsSalt` and `gramsSugar`? https://github.com/osu-mist/pizza-api/blob/cb99cca40ffe2d5e76f3cdd6236b597c2faa3132/openapi.yaml#L848-L856

jemisonf: Whoops, good catch

PR #15, opened 2019-12-03
CO-1659: adds unit tests for getDoughById

jemisonf:


src/tests/unit/doughs-dao.test.js
```diff
@@ -179,4 +184,52 @@ describe('test doughs dao', () => {
...
+    let result;
+    beforeEach(() => {
+      connectionStub = sinon.stub().returns(executeReturn);
+      doughsDao = proxyquireDoughsDao(
+        connectionStub,
+        serializeDoughsStub,
+        serializeDoughStub,
+      );
+    });
+    context('when called with an integer id', () => {
```
tsoliangwu0130: Looks like `context` is just an alias of `describe`. I'm fine using either one, but I think we should use the same one across all of the testings for the consistency.

src/tests/unit/doughs-dao.test.js
```diff
@@ -179,4 +184,52 @@ describe('test doughs dao', () => {
...
+      beforeEach(async () => {
+        result = doughsDao.getDoughById(1);
```
tsoliangwu0130: Again, ID must be a string. My bad I didn't catch it on the last PR, but you should modify your [DAO method](https://github.com/osu-mist/pizza-api/blob/6de21a4f650721e595469d986cb0a168cf3ae442/src/api/v1/db/oracledb/doughs-dao.js#L248-L264) to take string instead of a number. Test data should be modified as well.

src/tests/unit/doughs-dao.test.js
```diff
@@ -179,4 +184,52 @@ describe('test doughs dao', () => {
...
+
+      it('extracts rows from the return and passes them to the serializer', () => {
+        serializeDoughStub.should.have.been.calledWith(
+          executeReturnRows,
+          'doughs/1',
+        );
+      });
+
+      it('returns the result from the serializer', async () => result
+        .should.eventually.deep.equal(baseGetDoughsReturn));
```
tsoliangwu0130: I mentioned it on the yesterday code review but I guess you missed it and I didn't catch it on the second round code review lol. Again, data should be an "object" instead of a singleton array if we call get by ID method.

src/tests/unit/doughs-dao.test.js
```diff
@@ -179,4 +184,52 @@ describe('test doughs dao', () => {
...
+      context('when the database returns an empty result', () => {
+        beforeEach(async () => {
+          connectionStub.returns(emptyDatabaseReturn);
+          await doughsDao.getDoughById(1);
+        });
+
+        it('passes an empty array to the serializer', () => {
+          serializeDoughStub.should.have.been.calledWith(
+            [],
+            'doughs/1',
```
tsoliangwu0130: Think about it more (especially from BDD point of view), I don't think we should even pass the empty array to the serializer, instead, if we got an empty raw data from DAO, we should directly return 404 response. What do you think?

src/tests/unit/doughs-dao.test.js
```diff
@@ -179,4 +184,52 @@ describe('test doughs dao', () => {
...
+          );
+        });
+      });
+    });
+  });
```
tsoliangwu0130: You probably also wanna assert all number fields have been converted to integer in response. I also notice this branch is not based on the latest change from the last PR, so that the returned response in this PR is not correct (number fields are still strings). Can you merge the last PR to this branch?

src/tests/unit/doughs-dao.test.js
jemisonf: Technically, I think [this is the more idiomatic usage](https://lmws.net/describe-vs-context-in-rspec) of context and describe, and I was going to update future tests to do it this way. But I also don't feel that strongly about it and would be fine changing back if that would be easier to follow.

src/tests/unit/doughs-dao.test.js
jemisonf: I wrote this originally so that the logic for determining that empty result == 404 happens in the path code instead of the DAO code. I could change it to do something like throwing an exception, but in that case we'd have to do something like defining a custom exception so that the path code didn't return a 500 when the record wasn't found.

src/tests/unit/doughs-dao.test.js
jemisonf: That's a behavior that happens internally to the serializer, so I didn't think it made sense to test for it in this case. For the purpose of writing basically all of my DAO tests I've assumed that the serializer output is correct and the DAO doesn't touch the output of the serializer.

src/tests/unit/doughs-dao.test.js
jemisonf: Oops yeah I'll get this fixed, didn't updated the test logic and the docs after changing that in the openapi spec.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Good to know. I'll follow this in the future as well 👍

src/tests/unit/doughs-dao.test.js
jemisonf: Oh yeah, that's a problem. I will look into what's required to fix that.

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: Oh sorry for leaving the comment in the wrong place. I was mainly saying we should also have unit tests for serializer. Are you planning to work on it on the other ticket?

src/tests/unit/doughs-dao.test.js
tsoliangwu0130: That could work, or take a look at this [example](https://github.com/osu-mist/terms-api/blob/b9acb21bb42fa5a483af15f514ea9017f4366c9f/src/api/v1/db/oracledb/terms-dao.js#L50-L69).

jemisonf: Sounds good, I think that's basically what I'm going to go for.

src/tests/unit/doughs-dao.test.js
jemisonf: There are tests for the serializer but they need to be updated to actually test that they're converting into the string. I can open another ticket for that if you think that's the way to go.

tsoliangwu0130: Either way works for me. The reason I'm asking it is because the title of this PR sounds like it should cover all testing cases for `GET /doughs/{id}`.

jemisonf: That's a good point. I'm fine saying that's in scope for this ticket and doing that work in this PR.

src/tests/unit/doughs-serializer.test.js
```diff
@@ -99,6 +118,21 @@ describe('test doughs serializer', () => {
...
+          'bulkFermentTime',
+          'proofTime',
+          'gramsSugar',
+          'gramsOliveOil',
+        ].forEach((attribute) => serializedDoughs
+          .data[0].attributes[attribute].should.be.a('number'));
+      });
+
       it('returns an object with a data member that is an array', () => {
         serializedDoughs.data.should.be.a('array');
```
tsoliangwu0130: I might miss it, but I think you should also add the test case to assert it will return an object for get by ID response (serializedDoughs).

jemisonf: this is sort of implicitly tested for but I can make it explicit.

src/api/v1/paths/doughs/{doughId}.js
```diff
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+// import _ from 'lodash';
```
tsoliangwu0130: Can you remove it if you don't need this?

src/tests/unit/doughs-serializer.test.js
```diff
@@ -81,25 +81,17 @@ describe('test doughs serializer', () => {
...
-    describe('when it gets a raw dough with a valid format', () => {
+    describe('when it gets a singleton dough object with a valid format', () => {
       initializeSerializeDoughTest(emptyStringSpecialInstructionsDough, 'doughs');
 
       itCreatesTheCorrectTopLevelLink('/v1/doughs');
 
       itConvertsIntegerFieldsFromStrings();
 
+      it('returns `data` as an object', () => {
+        serializedDough.data.should.be.a('object');
```
tsoliangwu0130: Should be `an` object. Same mistake for the array assertion in `serializeDoughs`.

src/tests/unit/test-data.js
```diff
@@ -112,17 +112,17 @@ const postDoughsData = {
...
+    gramsWater: '400',
+    waterTemp: '90',
+    gramsYeast: '5',
+    gramsSalt: '15',
+    bulkFermentTime: '60',
+    proofTime: '15',
+    gramsSugar: '0',
+    gramsOliveOil: '0',
     specialInstructions: null,
   },
```
tsoliangwu0130: There are still some: `postIngredientData`, `normalizedIngredient`, `ingredientSerializerData`.

src/tests/unit/test-data.js
```diff
@@ -256,32 +264,32 @@ const postIngredientData = {
 const doughSerializerData = {
   nullSpecialInstructionsDough: {
     name: 'weeknight pizza dough',
-    id: 201,
-    gramsFlour: 500,
+    id: '201',
```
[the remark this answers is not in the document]
tsoliangwu0130: Also, at the `doughs-serializer.test.js` line 152, I think you don't need to convert the ID since it should be a string in the first place.

PR #16, opened 2019-12-06
CO-1660: implement and test update dough by ID

jemisonf:


openapi.yaml
```diff
@@ -889,11 +889,67 @@ definitions:
...
+                description: the time in minutes to spend on the bulk ferment
+                example: 60
+              proofTime:
+                type: integer
+                description: the time in minutes to spend on the proof
+                example: 15
+              specialInstructions:
+                type: string
+                description: any additional instructions not captured in the main object
+                example: keep the dough in the fridge during the bulk ferment
```
jemisonf: Added this because non-required values like `specialInstructions` that had a default value were getting populated with the default when they weren't included in `attributes`, leading them to get overwritten in every PATCH request if they'd previously been given a non default value. Open to suggestions if there's a more concise way to handle that issue.

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -265,4 +289,53 @@ const getDoughById = async (id) => {
...
+
+
+/**
+ * Update a specific dough recipe
+ *
+ * @param {object} body
+ * @returns {Promise<object>} the updated dough
+ */
+const updateDoughById = async (body) => {
+  if (_.isEmpty(body.data.attributes)) return getDoughById(body.data.id);
```
jemisonf: since I didn't see a way in openapi 2.0 to make `attributes` have at least one of the optional values, and jsonapi [requires](https://jsonapi.org/format/#crud-updating-resource-attributes) that "any" of the attributes can be included in the request, this handles the case where someone passes in an empty `attributes` object.

src/api/v1/paths/doughs/{doughId}.js
```diff
@@ -19,10 +20,24 @@ const get = async (req, res) => {
...
- * stub of patch function
+ * Update a dough with the ID specified in the request body
  *
- * @returns null
+ * @type {RequestHandler}
  */
-const patch = async () => null;
+const patch = async (req, res) => {
+  try {
+    if (req.params.doughId !== req.body.data.id) {
```
jemisonf: jsonapi [requires](https://jsonapi.org/format/#crud-updating-responses-409) this return and I didn't see a way to check this using the openapi spec

openapi.yaml
tsoliangwu0130: I'm confused. Why we need default values for returned attributes? Also, instead of using `minLength: 1`, why can't we put them in the `required` list?

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -265,4 +289,53 @@ const getDoughById = async (id) => {
...
+ *
+ * @param {object} body a PATCH request body
+ * @returns {object} `bindParams` and `query`
+ */
+const createPatchQueryAndBindParams = (body) => {
+  const inBindparams = getBindParams(body);
+
+  const query = doughsPatchQuery(inBindparams);
+
+  const bindParams = _.assign(
```
tsoliangwu0130: It's recommended using spread operation over assign according to the Airbnb JS style guide: https://github.com/airbnb/javascript#objects--rest-spread

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -265,4 +289,53 @@ const getDoughById = async (id) => {
...
 
-export { getDoughs, postDough, getDoughById };
+/**
+ * convert object body to an UPDATE query with correct bind params
+ *
+ * @param {object} body a PATCH request body
+ * @returns {object} `bindParams` and `query`
+ */
+const createPatchQueryAndBindParams = (body) => {
+  const inBindparams = getBindParams(body);
```
tsoliangwu0130: Should be `inBindParams`.

src/api/v1/db/oracledb/doughs-dao.js
tsoliangwu0130: You should define what's required attributes in the openapi file. If there are no required fields in the attributes, then `{}` should be a valid PATCH body in this case.

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -265,4 +289,53 @@ const getDoughById = async (id) => {
...
+
+  try {
+    const result = await connection.execute(
+      query,
+      bindParams,
+      { autoCommit: true },
+    );
+
+    const rawDough = convertOutBindsToRawDough(result.outBinds);
+    return serializeDough(rawDough, `doughs/${body.data.id}`);
```
tsoliangwu0130: If I were you, I'll probably add an additional logic check to make sure the returned ID is equivalent to the ID we pass from the PATCH body.

src/api/v1/paths/doughs/{doughId}.js
```diff
@@ -19,10 +20,24 @@ const get = async (req, res) => {
...
+      return errorBuilder(
+        res,
+        '409',
+        `ID ${req.params.doughId} in URL parameter does not match ID ${req.body.data.id} in body`,
```
tsoliangwu0130: Not important, but I think `req.body.data.id` and `req.params.doughId` can be destructured in the first place and be reused.

openapi.yaml
jemisonf: The default values were for the request body, not the body of the return value. They were used for some keys to default them to 0 or "" if they weren't required. If I'm remembering correctly, the reason for using minLength was that the database column is non null and we didn't want someone to be able to pass in an empty string as the value.

src/api/v1/db/oracledb/doughs-dao.js
jemisonf: Yeah, in this case because we can't require any single attribute, we have to make `{}` valid. This case doesn't report an error or anything, it's just a workaround for the fact that the query we use in the rest of the function doesn't work if it had no attributes to update.

src/api/v1/paths/doughs/{doughId}.js
jemisonf: That's true, although looking at it more I do like how as it is it's very clear where each id value comes from, and it might be less clear after destructuring. I don't feel super strongly about that though.

jemisonf:
> Got a 500 error with the following request:

@tsoliangwu0130 not 100% sure this is fixed because I don't think it reproduces on my computer, but I did add some fixes that might have solved the issue. Let me know if it's still cropping up.

jemisonf:
@tsoliangwu0130 thanks, I'm seeing it now

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -188,19 +211,20 @@ const getDoughs = async (filters) => {
...
  * @returns {object} bindparams
  * @throws {Error} when an attribute not in doughsProperties is found in body.
  * It's assumed that body has already been checked for invalid properties meaning that when
  * one is found here it's a server error rather than a user error.
  */
-const getPostBindParams = (body) => {
+const getBindParams = (body, baseValue = {}) => {
   const { attributes } = body.data;
-  const bindParams = doughsOutBindParams;
+  const bindParams = _.cloneDeep(baseValue);
```
jemisonf: making a shallow copy of `baseValue` meant we were modifying the original data structure, so when `postDoughs` was calling this with `doughsOutBindParams` the method was modifying `doughsOutBindParams` to also include the bind params for the original `POST` request, which would then show up in the `PATCH` request bind params, causing an error.

src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -114,12 +114,35 @@ const doughsGetQuery = (conditionals) => `SELECT ${doughColumnAliases} FROM DOUG
 
 /**
  * A query to create a new dough
+ *
  * @constant
  * @type {string}
  */
 const doughsPostQuery = `INSERT INTO DOUGHS (${doughColumns}) VALUES (${doughValues})
```
tsoliangwu0130: Just found there's a trailing space at the end of this line, also at line 141.

jemisonf: 👍 adding this as a rule in `.eslintrc`

PR #17, opened 2019-12-11
CO-1666: implement and test GET /ingredients/{ingredientID}

jemisonf:


src/tests/unit/ingredients-dao.test.js
```diff
@@ -239,4 +247,60 @@ describe('test ingredients dao', () => {
...
+      connectionStub.resetHistory();
+      serializeIngredientStub.resetHistory();
+      serializeIngredientStub.resetHistory();
+    });
+
+    context('when it gets an integer-formatted id', () => {
+      before(() => {
+        connectionStub = sinon.stub().returns(singleRecordReturn);
+        id = '1';
+      });
```
jemisonf: This is a stab at actually solving the DRY problem by initializing stubs and such in a `before` hook while assigning the result in a top level `beforeEach` hook. Replicates the (imo better) style I've used in [ginkgo](https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach) for BDD tests that uses `beforeEach` and `justBeforeEach`. Then `context` blocks can reassign stubs that are used specifically for that context, such as changing the database return, without having to duplicate logic, and we can minimize the amount of logic in each `it` block.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -196,4 +196,24 @@ const postIngredient = async (body) => {
...
+ */
+const getIngredientById = async (id) => {
+  const query = getIngredientsQuery('ID = :id');
+  const bindParams = { id };
+  const connection = await getConnection();
+  try {
+    const { rows } = await connection.execute(query, bindParams);
+    if (rows.length === 1) return serializeIngredient(rows[0], `ingredients/${id}`);
+    if (rows.length > 1) throw new Error("database return shouldn't have multiple rows");
+    return null;
```
Ruefa: I would rewrite this as
```
if (rows.length > 1) throw new Error("database return shouldn't have multiple rows");
if (rows.length === 1) {
  return serializeIngredient(rows[0], `ingredients/${id}`);
} else {
  return null;
}
```
This is makes the method more clear on what values it can return. You should strive to only have 1 return statement (same goes for exit conditions in loops) because it makes it clear where the method ends. We don't enforce this though and in this case I think it is fine to have 2.

src/api/v1/paths/ingredients/{ingredientId}.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Update an ingredient with ID `ingredientId` (stub)
+ *
+ * @type {RequestHandler}
+ */
+const patch = () => null;
+
+export {
+  get, patch,
+};
```
Ruefa: I don't think you need multiline exports until 4+ items. In any case make it consistent with your other exports.

src/tests/unit/ingredients-dao.test.js
```diff
@@ -5,10 +5,11 @@ import proxyquire from 'proxyquire';
 import sinon from 'sinon';
 import sinonChai from 'sinon-chai';
 
-import { getIngredientsData, postIngredientData } from 'tests/unit/test-data';
+import { getIngredientsData, postIngredientData, getIngredientByIdData } from 'tests/unit/test-data';
 import { openapi } from 'utils/load-openapi';
 import { GetFilterProcessor } from 'utils/process-get-filters';
 
+
```
Ruefa: remove extra line

src/tests/unit/ingredients-dao.test.js
Ruefa: Nice work! Looks real good.

src/api/v1/db/oracledb/ingredients-dao.js
jemisonf: I've [read the opposite](https://blog.timoxley.com/post/47041269194/avoid-else-return-early) in some cases but don't really have a strong take on it. I like how your suggested implementation makes the control flow really clear in this case though, and there's not really a risk that we'll end up even further nesting the `if`/`else` statements.

jemisonf: Hmm, eslint is marking that [as a linter error based on this rule](https://eslint.org/docs/rules/no-else-return). Given that, I think my inclination would be to do this:

```js
    if (rows.length > 1) throw new Error("database return shouldn't have multiple values");
    if (rows.length === 0) return null;
    return serializeDough(rows[0], `doughs/${id}`)
```

So it's checking the two "error" cases and then the return at the bottom is the "default" case.

Ruefa: Yeah the single point of exit is debatable. I lean toward having one return but its good to recognize points were multiple returns make more sense. The else was a mistake I like your solution.

PR #19, opened 2019-12-17
CO-1673 implement GET /pizzas/{pizzaId}

jemisonf:
Relationships!


src/tests/unit/test-data.js
```diff
@@ -452,6 +452,211 @@ const ingredientSerializerData = {
...
+          waterTemp: 90,
+          gramsYeast: 5,
+          gramsSalt: 15,
+          gramsSugar: 0,
+          gramsOliveOil: 0,
+          bulkFermentTime: 60,
+          proofTime: 15,
+          specialInstructions: '',
+        },
+        relationships: {},
```
jemisonf: based on [this](https://github.com/SeyZ/jsonapi-serializer/pull/209) PR, I think the fact that `relationships` shows up here is a bug in the serializer. It seems mostly benign though, since it doesn't show up in the API output.

src/api/v1/serializers/pizzas-serializer.js
```diff
@@ -0,0 +1,107 @@
...
+    transformFunction: transformRawPizza,
+  };
+
+  let options = serializerOptions(serializerArgs);
+  options = addCompoundRelationship(options, 'dough', doughResourceKeys);
+  options = addCompoundRelationship(options, 'ingredients', ingredientResourceKeys);
+
+  // need to depluralize type of `ingredients` compound resources . . . somehow
+  // returning `undefined` means it uses the default value
+  options.typeForAttribute = (attribute, data) => ('ingredientType' in data ? 'ingredient' : undefined);
```
jemisonf: Without this line, ingredients in `relationships` and `included` will have the type `ingredients` instead of `ingredient`. There may be a better way to do this than checking for `ingredientType` though.

src/api/v1/serializers/pizzas-serializer.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+    includedLinks: {
+      self: (collection, field) => resourcePathLink(
+        resourceUrl,
+        field[idField],
+      ),
+    },
+  };
+  return options;
+};
```
jemisonf: this should maybe get moved into the `jsonapi` util at some point, but I figured I would wait to do that until we had finished `GET /doughs` as well

src/tests/unit/pizzas-dao.test.js
```diff
@@ -0,0 +1,156 @@
...
+    let inputId;
+    before(() => {
+      serializePizzaStub = sinon.stub().returns(baseSerializerReturn);
+    });
+    beforeEach(async () => {
+      pizzasDao = proxyquirePizzasDao();
+      try {
+        result = await pizzasDao.getPizzaById(inputId);
+      } catch (err) {
+        resultError = err;
```
jemisonf: this workflow for handling errors, after many trials in previous test suites, is my favorite so far.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Fetch all the ingredients associated with the pizza with id
+ *  `pizzaId`
+ *
+ * @param {string} pizzaId
+ * @returns {Array} an array of unserialized ingredient objects
+ */
+const getPizzaIngredients = async (pizzaId) => {
+  const query = 'SELECT INGREDIENTS.ID AS "id", INGREDIENTS.TYPE AS "ingredientType", INGREDIENTS.NAME as "name", INGREDIENTS.NOTES AS "notes" FROM PIZZA_INGREDIENTS INNER JOIN INGREDIENTS ON PIZZA_INGREDIENTS.INGREDIENT_ID = INGREDIENTS.ID WHERE PIZZA_INGREDIENTS.PIZZA_ID = :id';
```
Ruefa: Let's split this bad boy up. I think you can use backticks ` for multiline strings.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param {string} doughId the ID of the dough to get
+ * @returns {object} a raw dough object
+ */
+const getPizzaDough = async (doughId) => {
+  const serializedDough = await getDoughById(doughId);
+
+  if (serializedDough === null) return {};
+
+  const deserializedDough = {
```
Ruefa: Seems like a waste to serialize the dough just to unserialize it. I believe there was talk about moving serialization into the daos but seeing this makes me question that. Seems less modular. Thoughts?

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,93 @@
...
+    id: serializedDough.data.id,
+    ...serializedDough.data.attributes,
+  };
+  return deserializedDough;
+};
+
+const getPizzaByIdQuery = `SELECT ${pizzaColumnAliases} FROM PIZZAS WHERE ID = :id`;
```
Ruefa: Globals should stay near the top of the file where they are more visible.

src/tests/unit/pizzas-dao.test.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+
+describe('test pizzas DAO', () => {
+  before(() => {
+    sinon.replace(config, 'get', () => ({ oracledb: {} }));
+  });
+  after(() => {
+    sinon.restore();
+  });
```
Ruefa: This can be simplified to `after(() => sinon.restore());`

Ruefa: The `before` can also be simplified that way but it might make the line to long. I can't tell with my eyes.

src/tests/unit/pizzas-dao.test.js
```diff
@@ -0,0 +1,156 @@
...
+      }
+    });
+    afterEach(() => {
+      serializePizzaStub.resetHistory();
+      connectionStub.resetHistory();
+    });
+    context('when it is called with a valid id', () => {
+      before(() => {
+        inputId = '1';
+      });
```
Ruefa: Any one line arrow function can be reduced like in the comment above.

src/tests/unit/test-data.js
Ruefa: You have to be real careful with odd bugs like this. They can create wonky edge cases that don't show up until a user finds it. At the very least you should understand why this is happening to actually determine its benignity.

src/api/v1/db/oracledb/pizzas-dao.js
jemisonf: I definitely think that having the path handler do serialization instead of the DAO makes sense in this case. This was the simplest option given the code base as it exists right now, but I'm totally in agreement that it's not an ideal solution.

src/tests/unit/pizzas-dao.test.js
jemisonf: eslint doesn't like it when you return assignments but I'm not sure if that's intended to apply in this case or not 🤔 .

[Rule here](https://eslint.org/docs/rules/no-return-assign)

src/tests/unit/pizzas-dao.test.js
jemisonf: looks like both are good to go 👍

tsoliangwu0130: Should this be `beforeEach` or just `before`? https://mochajs.org/#hooks

jemisonf: just `after`, this is to reset the configuration stub. The lower nested `context` blocks reset their sinon stubs individually.

src/tests/unit/test-data.js
jemisonf: That's fair, I was kind of at a loss about what to do about since it was showing up really consistently in my tests but wasn't causing issues in the actual API. I'm not sure what the best next steps would be to verify it won't cause issues down the road.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @param {string} pizzaId
+ * @returns {Array} an array of unserialized ingredient objects
+ */
+const getPizzaIngredients = async (pizzaId) => {
+  const query = `SELECT
+    INGREDIENTS.ID AS "id",
+      INGREDIENTS.TYPE AS "ingredientType",
+      INGREDIENTS.NAME AS "name",
+      INGREDIENTS.NOTES AS "notes"
```
tsoliangwu0130: Just a little nitpicking, I think the query here is not aligned and line breaking points are a little strange to me. It's not a big deal here, but it makes your query a bit hard to read if the query becomes larger, e.g. `getPizzaByIdQuery`. Take a look at [this file](https://github.sig.oregonstate.edu/ecs-data/students-api-contrib/blob/master/contrib.js) and if the way we organized queries can help a bit.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    const { rows } = await connection.execute(getPizzaByIdQuery, bindParams);
+
+    if (rows.length > 1) throw new Error('Returned multiple outputs for the same ID');
+    if (rows.length === 0) return null;
+
+    const { doughId } = rows[0];
+
+    const dough = await getPizzaDough(doughId);
+    const ingredients = await getPizzaIngredients(pizzaId);
```
tsoliangwu0130: It works, but you actually sent three SQL calls to get one pizza. It's not a big deal right now, but it will be a serious issue when you implement `GET /pizzas` endpoint. Can you try combining them into one SQL query?

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,112 @@
...
+ * @returns {Promise<object>} the full pizza record
+ */
+const getPizzaById = async (pizzaId) => {
+  const bindParams = { id: pizzaId };
+  const connection = await getConnection();
+  try {
+    const { rows } = await connection.execute(getPizzaByIdQuery, bindParams);
+
+    if (rows.length > 1) throw new Error('Returned multiple outputs for the same ID');
+    if (rows.length === 0) return null;
```
tsoliangwu0130: I got an unexpected 200 response with empty body if searching with `GET /pizzas/4`. Can you fix it?

src/tests/unit/test-data.js
tsoliangwu0130: `relationships: {},` this actually does show up in the API output (won't show for the first call). @jemisonf can you look into it and fix it?

src/api/v1/serializers/pizzas-serializer.js
```diff
@@ -0,0 +1,107 @@
...
+ * Add a compound relationship named `relationName` to `options`
+ * with keys `relationKeys` and primary key field `idField`
+ *
+ * @param {object} options
+ * @param {string} relationName
+ * @param {object} relationKeys
+ * @param {string} idField
+ * @returns {object} `options` with `relationName` added
+ */
+const addCompoundRelationship = (options, relationName, relationKeys, idField = 'id') => {
```
tsoliangwu0130: I might miss it, but will `idField` be overwritten by some other fields?

src/utils/jsonapi.js
```diff
@@ -36,7 +36,7 @@ const serializerOptions = (serializerArgs) => {
   const resourceUrl = resourcePathLink(apiBaseUrl, resourcePath);
   const options = {
     pluralizeType: false,
-    attributes: resourceKeys,
+    attributes: _.clone(resourceKeys),
```
jemisonf: this is a fix for the error where empty `relationship` keys were showing up in `included` resources. Because it was using a shallow copy of `resourceKeys`, subsequent calls to `addCompoundRelationship` in `pizzas-serializer` which mutated `attributes` were actually mutating `pizzaResourceKeys`, and the changes were persisting across API calls.

jemisonf: so I think the behavior where the serializer adds an extra `relationship` key _is_ still actually a bug, it's just that we can avoid it

src/api/v1/db/oracledb/pizzas-dao.js
jemisonf: I was concerned about the performance impacts having to join across four tables when fetching all of the pizza resources, so I was actually planning on asking about this. Do you think it's basically guaranteed that even one really expensive query is better than making a number of smaller queries? 

I'm also not sure what optimization strategies we would have available for either option. I had the thought that you could potentially cache the smaller queries on the server side, but I don't really know about what you can do to improve performance on the database side.

src/api/v1/serializers/pizzas-serializer.js
jemisonf: what do you mean?

tsoliangwu0130: So, from what I can see so far, this function was called in two places:

```js
options = addCompoundRelationship(options, 'dough', doughResourceKeys);
options = addCompoundRelationship(options, 'ingredients', ingredientResourceKeys);
```

My question was that since`idField` was not even get passed when you call this function, do we need to make it as a parameter?

src/api/v1/db/oracledb/pizzas-dao.js
jemisonf: I think I got it into a format that looks like what's used in the other file, let me know what you think

tsoliangwu0130: Much better. Make you modify other queries (include queries in `test-data.js`) as well.

jemisonf: Although a lot of the other ones are more procedurally generated and would be hard to split up. Was looking over the other files and they seem like they're in OK shape style wise too.

src/utils/jsonapi.js
jemisonf: nvm, changed this behavior to happen in `addCompoundRelationship` instead

src/api/v1/db/oracledb/pizzas-dao.js
tsoliangwu0130: Really a great question, I also chated with Jon (our DBA) about this topic.
To be honest, it really depends on how you destructure your query, but in this case and most of our cases, larger query will still be more efficient than multiple smaller queries. Caching can improve performance for sure, but you need to make sure the smaller queries used frequently enough so that they can be cached, otherwise it will only increase the cost.

jemisonf: Good point about caching. Going forward I'll switch to using a single query; I'm going to implement `includes` first using the current code and then I'll get the query updated.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -0,0 +1,162 @@
...
+  let doughsIncluded = false;
+  let head;
+  let ingredientsIncluded = false;
+  let index = 0;
+  const pizzas = [];
+
+  if (rows[0]) {
+    doughsIncluded = 'DOUGH_id' in rows[0];
+    ingredientsIncluded = 'INGREDIENT_id' in rows[0];
+  }
```
jemisonf: @tsoliangwu0130 follow up to our conversation earlier, here's what this looks like without the `included` parameter

jemisonf:
So thinking about performance, here's my thought:
* fetching `dough` relationships has no extra cost because the primary key is stored in the pizza record
* fetching `ingredient` relationships does have an extra cost because we need to join the `PIZZAS` table to the `PIZZA_INGREDIENTS` table, which would be true for any many-to-many relationships. It's less expensive than fetching the ingredient resources themselves though, since for that we have to join `PIZZA_INGREDIENTS` to `INGREDIENTS` in addition to the previous join.

So fetching just relationships is probably pretty fast, but I could see benefits for users from being able to just fetch the pizza resource without relationships, since that would be at least somewhat faster.

We might be able to use sparse fieldsets to get the best of both worlds by allowing users to request that we not get the relationships, but I don't know exactly how it works. That would also add a lot more work implementation work for me on this project which I'm not super excited about.

jemisonf:
> It can be a little confusing since an empty top-level included array can have two different meaning:

> 1. There are no related resources belong to the current resource
> 2. There might have some related resources, but included parameters are not provided

This is a serializer issue (not exactly a bug) where it leaves out `included` if it would be an empty array, even if the serialized object has relationship attributes. It's possible there's a way to fix it in the serializer configuration that I couldn't find.

src/tests/unit/test-data.js
```diff
@@ -561,17 +561,36 @@ const serializePizzaData = {
...
+  WHERE PIZZAS.ID = :id`,
+  getPizzaDoughQuery: `SELECT PIZZAS.ID AS "PIZZA_id", PIZZAS.DOUGH_ID AS "PIZZA_doughId", PIZZAS.NAME AS "PIZZA_name", PIZZAS.BAKE_TIME AS "PIZZA_bakeTime", PIZZAS.OVEN_TEMP AS "PIZZA_ovenTemp", PIZZAS.SPECIAL_INSTRUCTIONS AS "PIZZA_specialInstructions", DOUGHS.ID AS "DOUGH_id", DOUGHS.NAME AS "DOUGH_name", DOUGHS.GRAMS_FLOUR AS "DOUGH_gramsFlour", DOUGHS.GRAMS_WATER AS "DOUGH_gramsWater", DOUGHS.FLOUR_TYPE AS "DOUGH_flourType", DOUGHS.WATER_TEMP AS "DOUGH_waterTemp", DOUGHS.GRAMS_YEAST AS "DOUGH_gramsYeast", DOUGHS.GRAMS_SALT AS "DOUGH_gramsSalt", DOUGHS.GRAMS_SUGAR AS "DOUGH_gramsSugar", DOUGHS.GRAMS_OLIVE_OIL AS "DOUGH_gramsOliveOil", DOUGHS.BULK_FERMENT_TIME AS "DOUGH_bulkFermentTime", DOUGHS.PROOF_TIME AS "DOUGH_proofTime", DOUGHS.SPECIAL_INSTRUCTIONS AS "DOUGH_specialInstructions"
+  FROM PIZZAS 
+  LEFT JOIN DOUGHS ON PIZZAS.DOUGH_ID = DOUGHS.ID
+  
+  WHERE PIZZAS.ID = :id`,
+
+  getPizzaQuery: `SELECT PIZZAS.ID AS "PIZZA_id", PIZZAS.DOUGH_ID AS "PIZZA_doughId", PIZZAS.NAME AS "PIZZA_name", PIZZAS.BAKE_TIME AS "PIZZA_bakeTime", PIZZAS.OVEN_TEMP AS "PIZZA_ovenTemp", PIZZAS.SPECIAL_INSTRUCTIONS AS "PIZZA_specialInstructions"
+  FROM PIZZAS
```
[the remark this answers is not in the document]
jemisonf: Is this necessary for an automatically generated query? I understand the value of doing this for queries that are written in code by hand, but in this case we'd be changing our query generation logic just to make the unit test sample data look better.

openapi.yaml
```diff
@@ -134,6 +134,16 @@ paths:
         - pizza
       parameters:
         - $ref: '#/parameters/pizzaId'
+        - in: query
+          name: include
+          description: a comma separated list of relationship names to include in the response
+          type: array
+          items:
+            type: string
+            enum: [dough, ingredients]
```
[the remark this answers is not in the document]
jemisonf: One is a to-one relationship and one is a to-many relationship, so I was matching the terms used up to the endpoint names. I can change it if that's an issue.

Ruefa:
I don't have strong feelings on whether `included` should be included every time. I lean towards always returning it in the response because it is shown in the openapi documentation.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -14,96 +15,146 @@ const pizzaColumns = {
...
+        const ingredient = extractRawResource('INGREDIENT', ingredientsColumnNames, rows[index]);
+        if (ingredient) pizza.ingredients.push(ingredient);
+        index += 1;
+      }
+    } else {
+      index += 1;
+    }
+    pizzas.push(pizza);
+  }
+  return pizzas;
```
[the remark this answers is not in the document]
jemisonf: I pushed a change to update this while I think you were still reviewing the changes from the previous commit. I have a [comment here](https://github.com/osu-mist/pizza-api/pull/19#discussion_r359608373) about it, let me know what you think.

ianbrown9475:
@tsoliangwu0130 I agree with the comment about omitting `included` when the user doesn't make a request with the `include` parameter. This will make a distinction between not requesting included resources, and the resource just not having any.

Although it's possible to have a default value for `include`. In this case, `included` would still appear in the response with an array (possibly empty if there's not related resources). In this case, I don't think we can do anything about the ambiguity. Jsonapi 1.1 added the restriction that any time a user uses the `include` parameter, the response has to have `included`. So if we have an endpoint that by default responds with some included resources and a user requests `include=` (empty), we would still need to have `included: []` in the response, at least if we want to eventually be compliant with 1.1. I guess the user would know here that the array is empty because they specified no included resources, but it's just something to keep in mind.

@jemisonf we could use sparse fieldsets here, but it would be pretty clunky. If a client wants to exclude all relationships, they would need to specify a comma separated string that has every attribute of the primary resource like `fields[resourceName]=attribute1,attribute2,...`. If they want to exclude some relationships, they would need all attributes + the relationships they do want.

src/tests/unit/test-data.js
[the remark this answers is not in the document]
tsoliangwu0130: I guess yes. My reason is that even test data is also the code that we need to maintain in the future. It's not only to make the code look prettier, but also help others who maintain your code easier. That's also probably why I'm a little hesitated to write functions to generate most part of SQL queries automatically; it's kind of making the query implicitly to me and could potentially increase the difficulty of debugging in the future. 

When I first review your code, I felt like using functions to automatically generate query string was really smart, but I wasn't sure will there be any side effect. The reason I didn't bring this up was since nobody did this before and I was really curious how things work :)

openapi.yaml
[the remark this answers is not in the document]
tsoliangwu0130: Oh ok, but I guess the value in this array should be the resource of this API so they should be all plural. But you also brought a good point, so it's your call to decide which one you prefer 👍

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -14,96 +15,146 @@ const pizzaColumns = {
...
+const innerJoinIngredients = `LEFT JOIN PIZZA_INGREDIENTS ON PIZZAS.ID = PIZZA_INGREDIENTS.PIZZA_ID
+  LEFT JOIN INGREDIENTS ON INGREDIENTS.ID = PIZZA_INGREDIENTS.INGREDIENT_ID`;
 /**
- * Fetch all the ingredients associated with the pizza with id
- *  `pizzaId`
+ * A query to fetch a pizza with, optionally, its ingredients and doughs
  *
- * @param {string} pizzaId
- * @returns {Array} an array of unserialized ingredient objects
+ * @param {Array} included
```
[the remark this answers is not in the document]
jemisonf: Fixed this. If you're interested in more specific type annotations I was playing around with that [here](https://github.com/osu-mist/pizza-api/tree/test-types)

openapi.yaml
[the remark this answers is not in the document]
jemisonf: Hmm, that also makes sense. I don't really have any objection to that in general but leaving it as is the least work for me in this case 😆

src/tests/unit/test-data.js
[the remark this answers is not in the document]
jemisonf: That makes sense. One issue I've run into, on the subject of maintainability, is that when you're writing tests that check what queries were run against the database you have to do string comparisons on the queries, so differences in whitespace can cause the comparison to fail even if the query itself is actually the same. From that perspective, more readable SQL can be less maintainable since it can introduce errors that are harder to troubleshoot in the tests.

That said, I'm working on updating the style used in the SQL query right now to split up the aliases across multiple lines, I just wanted to mention this since that issue came up while I was working.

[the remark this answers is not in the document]
tsoliangwu0130: True, I also ran into this issue in my personal side project before. I recalled that I eventually used [dedent](https://docs.python.org/3.1/library/textwrap.html#textwrap.dedent) function (in Python, but I belive there should be something similar in JS) to solve the issue. So maybe you can look into this and see if that help?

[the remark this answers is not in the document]
jemisonf: Thanks, that's a really good suggestion. I'll look into that since that would solve the only real issue I had. My dream is a SQL parser that would let us actually compare the semantics of the two SQL statements but I don't think that exists, sadly.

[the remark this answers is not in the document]
tsoliangwu0130: well, I don't know will this module help, but looks like it's really popular and many people use it to build SQL query: https://knexjs.org/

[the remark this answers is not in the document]
jemisonf: Interesting, that might be something to look into. I was thinking something more like [this](https://www.npmjs.com/package/js-sql-parser) that can generate an AST from an existing query. But using a full on query generator might make sense too.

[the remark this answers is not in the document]
jemisonf: @tsoliangwu0130  got this updated after experimenting with a couple different strategies ([this](https://www.npmjs.com/package/flora) gets very close to being able to do AST diffs but throws an error on some valid oracle syntax), finally settled on just stripping extra whitespace out of the queries after having trouble getting `dedent` to work. Let me know what you think.

jemisonf:
Thinking about this more and I'm wondering if it would make sense to do some performance testing with `GET /pizzas` to see exactly what the performance cost of including relationship data by default is, since at this point we kind of just have to speculate about it. We could do something similar for other APIs and decide this on a case-by-case basis. Including relationship links by default seems reasonable to me that since we can do that for free. @tsoliangwu0130 @ianbrown9475 thoughts on that?

I would concur that having included be omitted when the parameter isn't passed and left as an empty array when there are no results is a reasonable idea.

src/utils/strip-whitespace.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Remove sequences of whitespace greater than 1 character from `str`
+ *
+ * @param {string} str
+ * @returns {string}
+ */
+const strip = (str) => str.replace(/\s+/g, ' ');
```
tsoliangwu0130: Well, I don't think it's correct logic to compare strings. You basically replace all continuous spaces to only one space, no matter where they are. It will make the following queries the same:

```
'select * from pizza where pizza_name = "Hello      Pizza"'
```

```
'select * from pizza where pizza_name = "Hello Pizza"'
```

tsoliangwu0130: Found this package should probably help: https://www.npmjs.com/package/dedent

src/api/v1/serializers/pizzas-serializer.js
```diff
@@ -53,25 +53,32 @@ const transformRawPizza = (rawPizza) => {
...
+  relationName,
+  relationKeys,
+  ignoreRelationshipData = false,
+) => {
   const idField = 'id';
   const resourceUrl = resourcePathLink(apiBaseUrl, relationName);
   options.attributes = _.concat(options.attributes, relationName);
   options[relationName] = {
-    ref: (collection, field) => field[idField],
+    ref: (collection, field) => (field && idField in field ? field[idField] : undefined),
```
tsoliangwu0130: I guess `idField in field` part is not necessary?

src/api/v1/serializers/pizzas-serializer.js
```diff
@@ -98,9 +105,13 @@ const serializePizza = (rawPizza, query) => {
   let options = serializerOptions(serializerArgs);
   if ('dough' in rawPizza) {
     options = addCompoundRelationship(options, 'dough', doughResourceKeys);
+  } else {
+    options = addCompoundRelationship(options, 'dough', doughResourceKeys, true);
   }
   if ('ingredients' in rawPizza) {
     options = addCompoundRelationship(options, 'ingredients', ingredientResourceKeys);
+  } else {
+    options = addCompoundRelationship(options, 'ingredients', ingredientResourceKeys, true);
```
tsoliangwu0130: The logic of line 106-110 is exactly the same as line 111-114. I guess it should be possible to combine together. The total line amount might not be improved a lot at this moment, but if we have more resources that can be included in the future or need to add more logic check here, it will be easier for you to extend/maintain.

src/api/v1/serializers/pizzas-serializer.js
jemisonf: yep, looks like it. Good catch.

src/utils/strip-whitespace.js
jemisonf: That's a good point, although I haven't ever actually used string literals with spaces in this code. It seemed like a reasonable thing to do at the time since after trying `dedent` and still having issues with strings not matching. I will revisit using dedent again though to see if I can fix it.

src/tests/unit/test-data.js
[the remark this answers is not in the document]
jemisonf: This is officially fixed and dedent is working.

src/utils/strip-whitespace.js
jemisonf: This is officially fixed and dedent is working.

PR #21, opened 2020-01-10
CO-1677: implement get pizzas

jemisonf:


.nvmrc
```diff
@@ -0,0 +1 @@
+v10.16.0
```
jemisonf: adding this so `nvm` can auto-detect the node version

src/tests/unit/pizzas-serializer.test.js
```diff
@@ -15,6 +15,55 @@ const {
...
+
+    if ('included' in serializedPizza && includedArgs.includes('dough')) {
+      serializedPizza.included.length.should.equal(1, 'it has only one element in `included`');
+
+      serializedPizza.included[0]
+        .should.deep.equal(goalSerializedPizza.included[0], 'it has a dough in `included`');
+    }
+  }
+};
+
```
jemisonf: moved common logic for checking if a resource was serialized correctly into this function

src/tests/unit/pizzas-dao.test.js
```diff
@@ -189,4 +204,114 @@ describe('test pizzas DAO', () => {
       });
     });
   });
+  context('getPizzas', () => {
```
Ruefa: This looks like one really long block of code. Use whitespace to break it up and make it more readable.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -64,13 +77,36 @@ const innerJoinIngredients = `LEFT JOIN PIZZA_INGREDIENTS ON PIZZAS.ID = PIZZA_I
...
+ *
+ * @param {object} query the query parameters passed to the HTTP request
+ * @returns {object} `bindParams` (object) and `conditionals` (string)
+ */
+const getConditionalsAndBindParams = (query) => {
+  const conditionalStatements = [];
+  const bindParams = {};
+  _.forEach(query, (parameterValue, parameterName) => {
+    if (pizzaFilters.includes(parameterName)
+      && parameterName.match(/filter\[.*\]/)) {
```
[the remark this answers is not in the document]
jemisonf: `pizzaFilters` is named badly, so it also includes parameters like `include` that don't match the regex. I think I'm going to add a `filter()` call when `pizzaFilters` is generated so that I can remove the check.

PR #22, opened 2020-01-15
CO-1683 implement post pizzas

jemisonf:


src/api/v1/db/oracledb/doughs-dao.js
```diff
@@ -31,14 +31,6 @@ const doughColumnNames = {
...
 };
 
-/**
- * convert an output bind param name like `nameOut` to a
- * property name like `name` by removing 'Out' at the end of the string
- * @param {string} outBindParamName
- * @returns {string} the property name
- */
-const outBindParamToPropertyName = (outBindParamName) => outBindParamName.slice(0, -3);
-
```
jemisonf: some minor refactoring here and in `ingredients-dao.js` to remove code that was duplicated between the three DAO modules

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -29,6 +34,32 @@ const pizzaColumns = {
...
+
+/**
+ * Query for updating a pizza record
+ *
+ * @const
+ */
+const pizzaPostQuery = dedent`INSERT INTO PIZZAS (NAME, BAKE_TIME, OVEN_TEMP, SPECIAL_INSTRUCTIONS)
+    VALUES (:name, :bakeTime, :ovenTemp, :specialInstructions)
+  RETURNING ID, NAME, OVEN_TEMP, BAKE_TIME, SPECIAL_INSTRUCTIONS
+  INTO :idOut, :nameOut, :ovenTempOut, :bakeTimeOut, :specialInstructionsOut`;
```
jemisonf: trying out writing queries by hand here instead of constructing them with code, especially since this one is totally static

src/utils/bind-params.js
```diff
@@ -26,4 +26,32 @@ const getBindParams = (body, allowedProperties, baseValue = {}) => {
...
+ * `serializeIngredient`.
+ *
+ * @param {object} outBinds
+ * @returns {object}
+ */
+const convertOutBindsToRawResource = (outBinds) => _.reduce(outBinds,
+  (rawResource, bindValueArray, bindName) => {
+    [rawResource[outBindParamToPropertyName(bindName)]] = bindValueArray;
+    return rawResource;
+  }, {});
```
ianbrown9475: If I'm understanding this right, you're taking an object `outBinds` and changing all of its key names using `outBindParamToPropertyName()`, right? If this is the case, I think [`_.mapKeys()`](https://lodash.com/docs/4.17.15#mapKeys) might be more helpful and simpler here.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -1,15 +1,19 @@
 import dedent from 'dedent';
 import _ from 'lodash';
 
+import OracleDB from 'oracledb';
```
ianbrown9475: Is there a reason for the blank line separating this import from the others above it?

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -20,6 +24,7 @@ const pizzaGetParameters = openapi.paths['/pizzas'].get.parameters;
 const pizzaFilters = _.map(pizzaGetParameters, (parameter) => parameter.name)
   .filter((parameterName) => parameterName.match(/^filter\[.*\]$/));
 
+
```
ianbrown9475: Could you remove this extra blank line for consistency?

src/api/v1/db/oracledb/pizzas-dao.js
ianbrown9475: `dedent` looks helpful, but I think the string might be a little more readable if you break the line after and before the backticks. Looks like the leading and trailing newlines will be stripped by `dedent`.
```js
const pizzaPostQuery = dedent`
  INSERT INTO PIZZAS (NAME, BAKE_TIME, OVEN_TEMP, SPECIAL_INSTRUCTIONS)
  VALUES (:name, :bakeTime, :ovenTemp, :specialInstructions)
  RETURNING ID, NAME, OVEN_TEMP, BAKE_TIME, SPECIAL_INSTRUCTIONS
  INTO :idOut, :nameOut, :ovenTempOut, :bakeTimeOut, :specialInstructionsOut
`;
```
What are your thoughts?

src/utils/bind-params.js
jemisonf: this is an excellent idea haha, I can't believe I didn't notice this

src/api/v1/db/oracledb/pizzas-dao.js
jemisonf: nope, that was an accident. possibly came up during a rebase and I didn't notice

src/api/v1/db/oracledb/pizzas-dao.js
jemisonf: fine by me, that makes sense

jemisonf:
> When I installed using npm 6.13.6, I get a slightly different lockfile. Looks like several dependencies have "optional": true added. Could you try rebuilding the lockfile with the latest npm version?
> When I query with name: "" I get a 500: Error: ORA-01400: cannot insert NULL into ("MISTSTU2"."PIZZAS"."NAME"). Could you look into this?

Fixed these. Looks like the pizza resource didn't get `minLength: 1` added to the required string values like the other resources did.

jemisonf:
> Could you add a TODO or something to postPizza about implementing some of the relationships functionality?

I actually misunderstood the spec here and forgot you can actually send relationship values in a `POST` request. Knowing now that you _are_ supposed to be able to do that, I think I'd prefer to add that in this PR. @ianbrown9475 does that seem reasonable?

src/utils/bind-params.js
jemisonf: Ah, now I understand why I used `reduce`. This:
```js
[rawResource[outBindParamToPropertyName(bindName)]]
```

Is actually destructing the single value array in `bindValueArray`, sort of like saying `rawResource[foo] = bindValueArray[0]`. So the fact that it's transforming keys and values means (I think) `reduce` is actually the best function to use here.

jemisonf: I am open to suggestions if it seems like there's a better way to do this, I think the fact that both of us got confused about what this function actually does is not a vote of confidence in its quality.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -217,4 +248,104 @@ const getPizzaById = async (pizzaId, query) => {
...
+  } catch (e) {
+    if ('errorNum' in e && e.errorNum === 2291) {
+      // oracleDB integrity constraint -- tried to insert invalid primary key for dough
+      return null;
+    }
+
+    throw new Error(`unhandled exception: ${e.message}`);
+  } finally {
+    connection.close();
+  }
```
jemisonf: steps here are basically:
1. check if ingredient IDs are valid
2. insert pizza resource
3. insert ingredient resources

We could merge these into one step if it was possible to write the query in such a way that we could get the ID of the pizza as we were creating it (I don't think this is possible) AND guarantee that all of the insertions would fail if the ingredient IDs used were invalid. `INSERT ALL` solves the second problem, but not the first, and using `connection.executeMany` could potentially solve the first problem but doesn't solve the second one.

I'm very interested to hear if there's a better way to solve this, but this is the best I could come up with.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -258,14 +230,40 @@ const updateIngredientById = async (body) => {
...
+ * @returns {boolean} returns true if ingredient IDs are valid
+ */
+const checkIngredientsExist = async (ingredientIds) => {
+  const connection = await getConnection();
+  const bindParams = _.reduce(ingredientIds,
+    (binds, curId) => {
+      binds[`id${curId}`] = curId;
+      return binds;
+    }, {});
+  const query = `SELECT Count(*) AS "count" FROM INGREDIENTS WHERE ID IN (${_.keys(bindParams).map((key) => `:${key}`)})`;
```
jemisonf: this query could return a false negative if someone passed a duplicate of the same relationship resource, like
```json
"data" : [
  { "type": "ingredient", "id": 1 },
  { "type": "ingredient", "id": 1 },
]
```

but I think that's supposed to be an error condition anyways so I'm not sure if it's an issue

src/api/v1/db/oracledb/setupDb.sql
```diff
@@ -51,7 +51,7 @@ CREATE TABLE "PIZZA_INGREDIENTS"
 	"DESCRIPTION" VARCHAR2(512 BYTE), 
 	 UNIQUE ("PIZZA_ID", "INGREDIENT_ID"),
 	 FOREIGN KEY ("PIZZA_ID") REFERENCES "PIZZAS" ("ID") ON DELETE CASCADE, 
-	 FOREIGN KEY ("PIZZA_ID") REFERENCES "INGREDIENTS" ("ID") ON DELETE CASCADE
+	 FOREIGN KEY ("INGREDIENT_ID") REFERENCES "INGREDIENTS" ("ID") ON DELETE CASCADE
```
jemisonf: extremely embarrassing error that had me blocked for like two days

src/utils/bind-params.js
ianbrown9475: Ah I see, good catch. Lodash also has a `_.transform()` function which basically works like reduce but you don't need to explicitly return the new result, making it better when transforming an object like this. As for the array destructuring, yeah it's a little easy to miss I think. Two options I can think of are

1. Add an intermediate variable
    ```js
    const [bindValue] = bindValueArray
    rawResource[outBindParamToPropertyName(bindName)] = bindValue;
    ```
2. just add a comment about how you're taking the first value in the array for its new value.

It seems to me that the array destructuring is more noticeable for a single variable name rather than surrounding a larger statement. Object destructuring doesn't seem to have this problem because we put spaces around the `{}`. Any other thoughts?

jemisonf: I think the first option is probably the best, since it lets us avoid using a comment. Adding in `transform` as well makes for a huge improvement:

```js
const convertOutBindsToRawResource = (outBinds) => _.transform(outBinds,
  (rawResource, bindValueArray, bindName) => {
    const [bindValue] = bindValueArray;
    rawResource[outBindParamToPropertyName(bindName)] = bindValue;
  }, {})
```

I'd agree that extra spacing makes object destructuring a little easier to read, although I think it's also just harder to use it in cases where it gets nested in an expression like this.

src/api/v1/db/oracledb/ingredients-dao.js
ianbrown9475: From our discussion, it looks like we'll want to handle a request like this with a 400. Hopefully this shouldn't be too difficult to implement as a general middleware.

jemisonf: @ianbrown9475 is there a good place to look for docs on adding middleware?

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -217,4 +248,104 @@ const getPizzaById = async (pizzaId, query) => {
...
+ * @param {object} body
+ * @returns {Promise<object>} the serialized, newly created pizza
+ */
+const postPizza = async (body) => {
+  const pizzaBindParams = getBindParams(body, pizzaProperties, pizzaOutBindParams);
+  const connection = await getConnection();
+  let ingredientQueries = [];
+  let ingredientBindParams = {};
+
+  const hasIngredients = 'relationships' in body.data && 'ingredients' in body.data.relationships;
```
ianbrown9475: You could use `_.has(body, 'data.relationships.ingredients);` here as well. This is nice because the undefined/null checking should be handled by the function.

src/api/v1/db/oracledb/pizzas-dao.js
ianbrown9475: I can't think of a better way to handle this. It should be fine to leave it how it is for now.

src/api/v1/db/oracledb/ingredients-dao.js
ianbrown9475: The [express-openapi](https://github.com/kogosoftwarellc/open-api/tree/master/packages/express-openapi#table-of-contents) table of contents has some details about different kinds of middleware and how they can be set up.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -29,6 +32,34 @@ const pizzaColumns = {
...
+    bindParam.type = properties.type === 'string' ? OracleDB.STRING : OracleDB.NUMBER;
+    bindParam.dir = OracleDB.BIND_OUT;
+    outBindParams[`${name}Out`] = bindParam;
+    return outBindParams;
+  }, { idOut: { type: OracleDB.NUMBER, dir: OracleDB.BIND_OUT } });
+
+/**
+ * Query for updating a pizza record
+ *
+ * @const
```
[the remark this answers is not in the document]
jemisonf: For me it has the semantics that the variable is initialized at start time and doesn't change, whereas `const` allows the variable to change as long as the underlying reference doesn't change (*[shakes tiny fist at javascript]*) so I think there's some value there. `@const` also shows up in the tooltips for vscode, whereas there's no indicator of if it uses a `const` or `let` declaration.

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -217,4 +248,104 @@ const getPizzaById = async (pizzaId, query) => {
   }
 };
 
-export { getPizzas, getPizzaById };
+/**
+ * Generate queries and bind params for an ingredients relationship resource collection
+ *
+ * @param {object[]} ingredientRelationshipData the `data` key of
+ * `body.data.relationships.ingredients`
+ * @returns {object[]} an object with `queries` and `bindParams` members
```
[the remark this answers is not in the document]
jemisonf: good idea, I'll add some `@throws` documentation

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -217,4 +248,104 @@ const getPizzaById = async (pizzaId, query) => {
...
+ */
+const getIngredientQueriesAndBindParams = async (ingredientRelationshipData) => {
+  const bindParams = {};
+  const ingredientIds = ingredientRelationshipData
+    .map((ingredient) => ingredient.id);
```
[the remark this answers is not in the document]
jemisonf: that exact approach does not work unfortunately since `ingredientRelationshipData` is an array and not an object. I don't see a way to use lodash methods for this functionality in a way that simplifies the operation being done here but let me know if you're aware of one.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -258,14 +230,40 @@ const updateIngredientById = async (body) => {
...
+  try {
+    const result = await connection.execute(query, bindParams);
+    if (parseInt(result.rows[0].count, 10) === ingredientIds.length) {
+      return true;
+    }
+    return false;
+  } finally {
+    connection.close();
+  }
+};
```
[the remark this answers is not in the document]
jemisonf: I'm interested in this approach, although I'm concerned it might be somewhat out of scope of this PR, which is already getting pretty huge. I'm wondering if it would make sense to defer this to the next PR, just to simplify the approval process for this one? 

I also would be curious to see what additional value this abstraction adds. I think that ensuring that every dao method uses `finally` is not a huge hurdle to overcome just with code reviews, and that this is definitely a jump in the complexity of the code that might not be super justified if that were the only benefit. Stuff I think we could consider adding here:

* some kind of error handling functionality that simplifies the interface for the request handler functions -- not sure exactly what this looks like
* automatic serialization of the result of the DAO function (is this possible?)
* validation of the serialized result, since we don't do this currently

And so on, curious to hear your thoughts here.

src/api/v1/db/oracledb/ingredients-dao.js
```diff
@@ -258,14 +230,40 @@ const updateIngredientById = async (body) => {
...
+      binds[`id${curId}`] = curId;
+      return binds;
+    }, {});
+  const query = `SELECT Count(*) AS "count" FROM INGREDIENTS WHERE ID IN (${_.keys(bindParams).map((key) => `:${key}`)})`;
+  try {
+    const result = await connection.execute(query, bindParams);
+    if (parseInt(result.rows[0].count, 10) === ingredientIds.length) {
+      return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
jemisonf: this and the comment above are good ideas but also a good reminder that I should add some unit tests to this function before I start trying to refactor it

src/api/v1/db/oracledb/pizzas-dao.js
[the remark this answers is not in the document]
ianbrown9475: Oh I see. In that case, I think you could do `_.map(ingredientRelationshipData, 'id')`

src/api/v1/db/oracledb/pizzas-dao.js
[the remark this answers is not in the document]
ianbrown9475: Since this function should only return a promise, I'm not sure if we want `@throws` since it won't actually throw any exceptions, just return rejected promises.

src/api/v1/db/oracledb/ingredients-dao.js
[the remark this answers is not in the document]
ianbrown9475: From our meeting, this should be a good investigation to do in a separate ticket. We should look at similar usage in other apis, especially the students api. If all goes well with changes to the pizza api, we can bring in the changes to the skeleton.

src/api/v1/db/oracledb/pizzas-dao.js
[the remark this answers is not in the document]
jemisonf: yep, that checks out. It would be nice if jsdoc had something like an `@resolves` and `@rejects` tag for promises.

src/api/v1/db/oracledb/ingredients-dao.js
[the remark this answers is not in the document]
jemisonf: Sounds like a plan. I will get at ticket opened in a sec, and I would like to plan on implementing this in the next instead instead of this one.

jemisonf:
> I'm assuming you'll just want to add a pattern or something to this in the openapi doc

yeah this is how we solved that with other resources, I'll add it really quick

jemisonf:
^ just pushed a commit to fix an issue with the relationship checking middleware where it wasn't handling requests with no body correctly

PR #24, opened 2020-01-31
CO-1686: implement patch pizza

jemisonf:


src/tests/unit/pizzas-dao.test.js
```diff
@@ -6,14 +6,17 @@ import proxyquire from 'proxyquire';
...
+  getPizzasData, getPizzaByIdData, postPizzaData, updatePizzaByIdData,
+} from './test-data';
 
 const should = chai.should();
 chai.use(chaiAsPromised);
 chai.use(sinonChai);
 
 let checkIngredientsExistsStub;
-let connectionStub;
+let executeStub;
```
jemisonf: this has a huge diff because `connectionStub` was renamed to `executeStub`; real changes are at the bottom of the file

src/utils/dao-errors.js
```diff
@@ -0,0 +1,38 @@
...
+   */
+  constructor(resource, id) {
+    this.id = id;
+    this.resource = resource;
+  }
+
+  get [Symbol.toStringTag]() {
+    return 'ResourceNotFoundError';
+  }
+}
```
jemisonf: this is the simplest way to make it possible to check error types with Chai's `should.be.a(...)` that I know of

src/utils/dao-errors.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable max-classes-per-file */
+/* eslint-disable class-methods-use-this */
```
jemisonf: not terribly happy about disabling eslint errors but:
* disabling `max-classes-per-file` means we don't have to write one file per error type (bleh)
* `class-methods-use-this` appears to literally break `Symbol.toStringTag` which is required for use with Chai in our tests

src/utils/with-connection.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ *
+ * @param {withConnectionClosure} closure must be an async function
+ * @param {object[]} closureArgs arguments to pass to `closure`
+ * @returns {*} the result of calling `closure` with a new `connection` object
+ */
+const withConnection = async (closure, ...closureArgs) => {
+  const connection = await getConnection();
+  return closure(connection, ...closureArgs).finally(() => connection.close());
```
jemisonf: Trying to decide if `closureArgs` is premature abstraction or not (it's not currently used by any caller in `pizzas-dao`). Would you prefer this:
```js
const doThing = (connection, query) => await connection.execute(query)

const foo = () => {
  await withConnection(doThing, "SELECT * FROM PIZZAS");
}
```

Or this:
```js
const doThing = (query) => {
  return withConnection((connection) => {
    return await connection.execute(query)
  });
};

const foo = () => {
  await doThing("SELECT * FROM PIZZAS")
}
```

I strongly suspect we'll want to do one of these at some point, because I'm noticing already that there's some duplicate work between `updatePizzaById` and `getPizzaById` in the case where `updatePizzaById` gets a body with no attributes.

src/utils/dao-errors.js
```diff
@@ -0,0 +1,38 @@
...
+
+class ResourceNotFoundError {
```
jemisonf: I felt like it made more sense to define these errors as they came up, rather than trying to come up with an exhaustive list now.

src/api/v1/paths/pizzas/{pizzaId}.js
```diff
@@ -9,20 +10,45 @@ import { errorBuilder, errorHandler } from 'errors/errors';
...
-const patch = () => null;
+const patch = async (req, res) => {
+  try {
+    if (req.query.pizzaId !== req.body.data.id) {
+      return errorBuilder(
+        res,
+        '409',
+        `ID ${req.params.pizzaId} in URL does not match ID ${req.body.data.id}`,
+      );
+    }
```
ianbrown9475: I seem to be getting this message even when the two IDs are the same: `"detail": "ID 1 in URL does not match ID 1"`

src/utils/with-connection.js
ianbrown9475: I think using `closureArgs` should be ok since it should help simplify the code like in your example.

ianbrown9475 (changes requested):

src/utils/with-connection.js
jemisonf: Cool, I'll keep it in. Appreciate the sanity check on that.

src/utils/dao-errors.js
jemisonf: Nice, that looks like a better solution

src/api/v1/db/oracledb/pizzas-dao.js
```diff
@@ -348,4 +365,151 @@ const postPizza = async (body) => {
...
+ * @param {OracleDB.connection} connection
+ */
+const replaceIngredients = async (
+  pizzaId, ingredientsQueries, ingredientsBindParams, connection,
+) => {
+  ingredientsBindParams.pizzaId = pizzaId;
+  const deletePizzaIngredientsQuery = dedent`
+    DELETE FROM PIZZA_INGREDIENTS WHERE PIZZA_ID = :pizzaId
+  `;
+  await connection.execute(deletePizzaIngredientsQuery, { pizzaId }, { autoCommit: true });
```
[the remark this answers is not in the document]
jemisonf: I'm open to this change. My concerns are:
* in theory, we've already validated the ingredient IDs and can guarantee that they're safe
* I've had bad experiences with not committing changes, including a situation where what I think happened was that a connection made uncommitted changes, then crashed, and then held a lock on on the resources it changed for like ~24 hours and made it impossible to update them

Thoughts on that? Given the combo of 1 and 2 I think the current design is _relatively_ safe. But partly that's because I have very little confidence around working with oracledb commits.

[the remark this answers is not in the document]
ianbrown9475: My bad, I didn't mean that there shouldn't be a commit, just that the commit should only occur in the last statement (the insertIngredients one). The main concern is that if the deleteIngredients statement succeeds, is committed, then the insertIngredients statement fails, I'm pretty sure this would end up just deleting the ingredients for a pizza when a patch was requested. We would want to respond with an error like a 500 here, but as a user, I would expect nothing to have changed.

So I think if we execute the first statement without committing, then only commit after the second statement, this will be prevented. Then, if an error occurred in the second statement after a successful first statement, the `finally` block would close the connection and [rollback any uncommitted changes](https://oracle.github.io/node-oracledb/doc/api.html#connectionclose).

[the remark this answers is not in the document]
jemisonf: > Then, if an error occurred in the second statement after a successful first statement, the finally block would close the connection and rollback any uncommitted changes.

Cool, this basically answers my main concern. I'll push up the changes now.

ianbrown9475 (changes requested):
Could you update the openapi doc to handle unknown properties? Our general approach is to allow unknown query parameters but reject unknown properties on request bodies. For example, when I include "key": "val" in `data.attributes`, I get a 500: `Error: Invalid attribute key found`

ianbrown9475 (approved):

